PR #22, opened 2021-06-03
WIP: Uses buildx for multi arch builds

rverma-nsl:
1. Bumped up to go 1.16 to support mac m1 too for local testing, can safely avoided


ArchiFleKs:
Hi, any update on this, would love to be able to use this on graviton instances

simonmarty:
We recommend that if you try [developing inside an AL2 container](https://code.visualstudio.com/docs/remote/containers) if you want to work on this project locally on your mac.

bjoernw:
This would be needed to support graviton instances, too, which AWS is promoting quite a bit.

PR #43, opened 2021-08-23
Add support for Go templating

martinhaus:
*Description of changes:*

We're trying to use the AWS CSI secrets provider as a universal way of integrating AWS Secrets Manager to all of our apps that we run in EKS.
Ultimately we would like to be able to specify a format in which the file containing secrets from SM should be created. For example, we would like to be able to format the secrets in the following fashion, in order to be able to source them in bash later:
export KEY1=VALUE1
export KEY2=VALUE2.

A similar functionality already exists in Vault Agent used with Hashicorp's Vault 
https://www.vaultproject.io/docs/platform/k8s/injector#secret-templates

This PRs adds an option to specify a Go Template in a objectTemplate variable in the object's parameters. This template is then used to transform the JSON format that is produced by default into a desired format.

---


martinhaus:
@lasred please take a look when you have time

tiparega:
Hi,

I had the same problem and discovered some things not working on @martinhaus code, so I made another PR. Work is still needed for testing and code improvement, as I don't know Golang. Please, check PR

yrsurya:
Is there any ETA when this feature can be available?

joebaro:
There are three existing options that can already be used to achieve what you want:

1. Secrets manager does not require a specific format for the secrets. You can store the secret in any format you wish and that is how it will appear in the mounted file. If you are using the console you can use the plain text tab to store the secret as you wish.
2. The AWS plugin for the CSI driver supports a [jmesPath](https://github.com/aws/secrets-store-csi-driver-provider-aws#secretproviderclass-options) JSON parsing option which will parse the secret and mount the parts you specify separately.
3. In conjunction with the jmesPath option above, you can use the CSI driver to [sync the secrets as K8s secrets](https://secrets-store-csi-driver.sigs.k8s.io/topics/sync-as-kubernetes-secret.html) and set them as [shell variables](https://secrets-store-csi-driver.sigs.k8s.io/topics/set-as-env-var.html) directly so your shell script does not have to parse the files. Note you will also have to start the CSI driver with ```--set syncSecret.enabled="true"```.

yrsurya:
@joebaro  we are looking to see if this can support what Vault can do similar to here - https://developer.hashicorp.com/vault/docs/platform/k8s/injector#secret-templates

```
{{ with secret "/path/to/secret" }}
    {{ range $k, $v := .Data }}
        {{ $k }}: {{ $v }}
    {{ end }}
{{ end }}
```

joebaro:
The CSI driver uses a daemon set architecture. That is to say the driver and plugins run as trusted containers on every node. This means the introduction of an executable template raises additional security concerns around the ability to break out of or misconfiguration of the templates.

It is not clear what format your secrets are in or if you are moving them over and trying to reformat them. Option 1. may still apply to your use case, you could store them directly in the desired format. Secrets Manager and the CSI driver pass through the secrets unchanged. 

The vault solution you pointed out is not the CSI vault plugin but an alternate sidecar implementation. There are also other sidecar solutions like [External Secrets](https://external-secrets.io) which work with Secrets Manager and [provide templates](https://external-secrets.io/v0.8.5/guides/templating/) that may or may not be a better fit for your use case.

Obviously there is also the option of cloning this repo and running your own private build to add on the features you are looking for. Instructions for that are provided in the [Private Builds](https://github.com/aws/secrets-store-csi-driver-provider-aws/tree/52ce2ef4ee74154cfe89cfd8c570052e4efbb990#private-builds) section of the readme.

Issue #15, opened 2021-05-21
feature request: helm chart for installation


vito-laurenza-zocdoc:
Providing a helm chart for installation would make it easier for folks to customize the installation properties.

joebaro:
Thank you for your feedback. We have noted this as a feature request.

pierluigilenoci:
I support the request

pierluigilenoci:
@joebaro is it possible to prioritize this feature given the interest of many people and the fact that it would speed up adoption by many? Thanks! ❤️

pierluigilenoci:
@vito-laurenza-zocdoc I took the liberty of creating one: #35

pierluigilenoci:
I did the PR but for now no sign of life from the maintainers. Will they be on vacation?

pierluigilenoci:
@vito-laurenza-zocdoc you can find the chart here:
https://github.com/aws/eks-charts/tree/master/stable/csi-secrets-store-provider-aws

Igor992:
@pierluigilenoci thanks for this chart, is it possible to support **annotations** here - https://github.com/aws/eks-charts/blob/master/stable/csi-secrets-store-provider-aws/templates/serviceaccount.yaml#L4

Scenario is that I need to use custom template from terraform to apply **eks.amazonaws.com/role-arn** into custom service account, but that fails because initial manifest doesn't support that option.

Docs: https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html

Thanks!

pierluigilenoci:
@Igor992 this can certainly be done. You can do a PR on the EKS Chart repo to get this feature.

If I remember correctly, however, AWS is developing a new version of the chart from scratch so I don't know how much it is worth doing. I no longer find the discussion about it.

Igor992:
@pierluigilenoci Do we know time when this new chart will be available?

pierluigilenoci:
You have to ask the AWS team that takes care of it.
I have no idea.

pierluigilenoci:
[...]

PR #35, opened 2021-07-09
Helm Chart for the csi-secrets-store-provider-aws provider

pierluigilenoci:
*Issue #, if available: #15

*Description of changes: Added a preliminary version of the chart that installs the software with the same configuration as the [deployment files](deployment/aws-provider-installer.yam) already present on the repo. Feel free to improve it or to suggest changes.

I put my name as maintainer because I did not know who else to insert but more than willingly I pass the burden and honors to someone else.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

This chart is a basic chart to replace point 3 of this guide: https://aws.amazon.com/fr/blogs/security/how-to-use-aws-secrets-configuration-provider-with-kubernetes-secrets-store-csi-driver/

PR #56, opened 2021-10-18
Added helm chart using existing yaml

jordiprats:
Fixes issue #15

Using the existing yaml file I have created a helm chart so that there's no need to maintain two separate versions of the same yaml. I can also submit an github action to release this helm chart as you do on other AWS repos if this approach gets accepted

### DIFF
--- a/deployment/secrets-store-aws-provider-installer/.helmignore
+++ b/deployment/secrets-store-aws-provider-installer/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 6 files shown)


pierluigilenoci:
@jordiprats have you seen my PR? #35

pierluigilenoci:
@jordiprats the chart is here: https://github.com/aws/eks-charts/tree/master/stable/csi-secrets-store-provider-aws

simonmarty:
We're in the process of writing a helm chart & pod security policy for this repository, in the meantime you can use the one referred above.

pierluigilenoci:
@simonmarty  when I tried to create the chart in this repo I was invited to put it in the EKS Chart one. I am a little confused.

Issue #10, opened 2021-05-05
Support arm64


rverma-nsl:
Prefer multi arch builds using buildx

joebaro:
Thank you for your feedback. We have noted this as a feature request.

mohandass-pat:
Hi, any update on this one?

gbhosal:
I'm also waiting for this feature. Any update on this?

rverma-nsl:
Feel free to use docker pull ghcr.io/nslhb/secrets-store-csi-driver-provider-aws:latest, while this came in mainline.

lpmatos:
@rverma-nsl thx!

PR #69, opened 2022-02-10
Perform multi-arch builds including arm64

otterley:
Fixes: #10 

Use `docker buildx build` to create multi-architecture images supporting both the amd64 (x86-64) and arm64 (armv8) platforms. 

As a side benefit, improves local build times (`make build`) for developers by rebuilding only if any of the source files has changed.

Note that Docker build times will increase due to software emulation of the foreign architecture. This isn't the only way to perform multi-arch builds; but it is the best way to do it using a single build host, without requiring access to a Docker builder running on each target having the desired architecture.

Tested successfully using a private Amazon ECR repository.

In addition, I've committed `go.sum` as [best practice dictates](https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file).

---


erikwilson-rc:
@StevenEmelander @lasrer @aramase are there any plans for an r3 release with arm64 support? it would be nice to see this go in! :)

smobyrne:
ARM64 images are now available. Closing in favor of #78

PR #144, opened 2022-11-04
add support for priorityClassName

JRemitz:
**Issue #:** N/A

**Description of changes:**
This will just add the ability to configure a [`priorityClassName`](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass) for pod scheduling priority. We're seeing contention with Karpenter when daemonsets are getting scheduled and having enough resources at that time to schedule all other workloads.  Secrets Store CSI should have priority.


simonmarty:
Merged in #171

Issue #92, opened 2022-05-16
EKS K8s secret cannot be created from volume mount with secret store csi driver


arimaverick:
I want to pass the aws secrets manager secret as an environment variable to the eks container. However even after correctly volume mounted the secret, the kubernetes  secret could not be created from the volume mount.  
I am using the roles and service account mentioned in the document.

**To Reproduce**
Here is my secretprovider class:
```
apiVersion: secrets-store.csi.x-k8s.io/v1
kind: SecretProviderClass
metadata:
  name: aws-secrets
spec:
  provider: aws
  secretObjects:
  - secretName: newsecret     # name of the Kubernetes Secret object
    type: Opaque         # type of the Kubernetes Secret object e.g. Opaque, kubernetes.io/tls
    data:
    - objectName: dbpass   # name of the mounted content to sync. this could be the object name or the object alias 
      key: password      # data field to populate
  parameters:                    # provider-specific parameters
    objects:  |
      - objectName: dummysecret
        objectType: secretsmanager
        objectAlias: dbpass
```

My deployment manifest section where I am passing the secret as an Environment variable:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ghost
  labels:
    app: ghost
spec:
  selector:
    matchLabels:
      app: ghost
      tier: frontend
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app: ghost
        tier: frontend
    spec:
      serviceAccountName: ghost-sa
      containers:
      - image: ghost:1-alpine
        name: ghost
        env:
        - name: database_client
          value: mysql
        - name: database_connection_host
          value: XXXXXX
        - name: database_connection_user
          value: ghostadmin
        - name: database_connection_password
          valueFrom:
            secretKeyRef:
              name: newsecret
              key: password
        - name: database_connection_database
          value: ghostdb1
        ports:
        - containerPort: 2368
          name: ghost
        volumeMounts:
        - name: ghost-persistent-storage
          mountPath: /var/lib/ghost/content
        - name: ghostdb-pass
          mountPath: /mnt/ghostdb-pass
          readOnly: true
      volumes:
      - name: ghost-persistent-storage
        persistentVolumeClaim:
          claimName: efs-storage-claim
      - name: ghostdb-pass
        csi:
          driver: secrets-store.csi.k8s.io
          readOnly: true
          volumeAttributes: 
            secretProviderClass: aws-secrets
```

[...]

PR #162, opened 2022-12-20
Add secret permissions to clusterrole, annotations to Helm chart.

reflection:
*Issue #, if available:*
#92

*Description of changes:*
- Add secrets permissions to clusterrole, needed if creating secretObjects in SecretProviderClass
- Allow modifications to service account annotations (i.e. linking IRSA role arn)


charts/secrets-store-csi-driver-provider-aws/templates/rbac.yaml
```diff
@@ -41,4 +49,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
 {{ include "provider.labels" . | indent 4 }}
+  {{- with .Values.rbac.serviceAccount.annotations }}
+  annotations:
```
simonmarty, line 53: Curious, what's your use case for needing [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)?

simonmarty (changes requested):

charts/secrets-store-csi-driver-provider-aws/templates/rbac.yaml
reflection: @simonmarty sorry for the delay in response, we link the role-arn via an `rbac.serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn` annotation (here's the terraform code with the forked helm chart):

```
# AWS helm chart missing required serviceaccount iam role annotation and clusterrole secret permissions
# https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/162
resource "helm_release" "secrets_store_csi_driver_aws" {
  name      = "secrets-store-csi-driver-provider-aws"
  chart     = "../../../atg-xdr-helm/environments/secrets-store-csi-driver-provider-aws"
  namespace = var.namespace

  set {
    name  = "rbac.serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn"
    value = module.secrets_manager_role.iam_role_arn
  }
}
```

simonmarty (approved):

s-pouya-s:
@danmancuso @joebaro 
please somebody approves this PR

theboyisno-one:
> @danmancuso @joebaro please somebody approves this PR

plz approve and merge @joebaro @danmancuso @simonmarty

joebaro (commented):
There are a few problems here. First I can not reproduce the original issue (more on that below). Second, it is not the AWS plugin that updates the K8s secrets, that is done by the CSI driver itself running under a different service account. Also, the CSI driver is already setting up [these permission on their role](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/templates/role-syncsecret.yaml) when sync is turned on. We also would not set these permission globally for all installations as they don't all use K8s secrets (which is why the CSI driver does them conditionally).

I suspect what may have happened is an older version of the CSI driver was not setting the role permission correctly on upgrade or something was set by hand interfering with it. When I upgrade the sync option on the latest CSI driver I do see the cluster role setup correctly:
```
$ helm install -n kube-system secrets-provider-aws aws-secrets-manager/secrets-store-csi-driver-provider-aws
$ kubectl -o yaml -n kube-system get clusterroles secretprovidersyncing-role
Error from server (NotFound): clusterroles.rbac.authorization.k8s.io "secretprovidersyncing-role" not found

$ helm upgrade -n kube-system csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver --set syncSecret.enabled=true
$ kubectl -o yaml -n kube-system get clusterroles secretprovidersyncing-role
    ...
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
```

As to the annotation on the service account, this is the service account for the AWS provider daemonset, there is no IAM role associated with it. The role association and annotation should be on the application SA, and this is typically done for each application pod using `eksctl create iamserviceaccount`.

PR #163, opened 2022-12-22
Template secrets and joins

tiparega:
*Issue #43*

*Description of changes:*
I added few changes over @martinhaus commits.

- I merged current master branch
- One is to not override main secret, as that one in json format is used if no change is detected (specially interesting for sync).
- The other is I added the option to join all templated secrets into one file. I needed because I was mixing some secrets into a single YAML file.

It's properly working, but I don't know Golang and I didn't changed tests, so code can surely be improved.


yrsurya:
is there any ETA when can we get this feature available?

PR #171, opened 2023-01-19
[helm] Add priorityClassName

maur1th:
*Description of changes:*

We sometimes have pods failing on new nodes because they get scheduled before the aws provider csi driver pod. The `secrets-store-csi` Helm chart [supports `priorityClassName`](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/values.yaml#L90) to prevent this kind of issue from happening. This PR implements the same mechanism here.


simonmarty (approved):

kitos9112:
I think this is kind of essential in any EKS deployment where the pod limit per worker node applies. Is there any way for getting this prioritised?

danmancuso (approved):

PR #172, opened 2023-01-20
Make sure to register cannotAssumeRoleWithWebIdentity as fatal

nclaeys:
*Description of changes:*

Fixed a regression issue where certain errors are not registered as fatal.
When trying to get parameters from ssm with an aws role with an incorrect trust relationship, we now get a very generic error:

`Failed to fetch parameters from all regions`

In older versions of the csi driver, we would get a very clear error message which was actionable for us and the users of the platform:

`Not authorized to perform sts:AssumeRoleWithWebIdentity`

The changes in this PR will make sure we show the clear error message again while not breaking any of the existing functionality.


goyalya:
Thank you for requesting the review of Pull Request. We are looking into it.

nclaeys:
@goyalya Is there an issue with the PR or another reason why it is not approved yet?

mandydas (approved):

Issue #131, opened 2022-09-29
Windows nodes support or workaround?


ibaki-p:
We are deploy a mix of Linux & Windows containers in an EKS cluster. Although the Kubernetes Secrets Store CSI Driver supports both Linux & Windows. This AWS provider does not seem to support Windows. Is that the case? Any workaround you can think of? Thanks

jbct:
Hi ibaki-p. Could we get some more information on your deployment configuration?

ibaki-p:
Hi all, sorry about the delay. We decided to migrate our workloads to run on Linux only. If it can help, at the time I created this issue, we were deploying an EKS cluster that had both Linux managed nodes and Windows Self hosted nodes. The workloads running on these nodes (a mix of Windows and Linux containers of course) required access to secrets in AWS Secret Manager. We were therefore using the AWS Secrets Store CSI driver to mount secrets. It was working all well and good for the Linux workloads but not for the Windows workloads.

saedx1:
@jbct I actually have a PR up to add support. Not much work once you figure out what's different for windows (just some path separator thing and closing the file before renaming it). I did add a Dockerfile and added it to the deployment/- installer. It passes all the tests, and I've tested it on EKS with a windows node, 6 different deployments, and mounted secrets as well as used them as env vars.

I'd appreciate if it gets merged and if we get a public windows docker image on ecr.

jbct:
Thanks for the context everyone, and thanks Saed for the PR. At the moment, Secrets Manager doesn't plan on supporting Windows explicitly. However, you might be able to use the flag `driver-writes-secrets` which bypasses the plugin and may solve the issues you're encountering on the Windows platform. I'll leave this issue open as a feature request while we discuss future opportunities to expand support for the plugin.

saedx1:
@jbct  Can you elaborate more on this, please? Won't you still need it to be running on the windows node regardless of who ends up writing?

mhatreas:
@jbct  We are also facing similar challenges in using AWS Secrets Manager or AWS parameter store with EKS for our Windows PODs running on Windows worker node.
We have raised Support case too  with AWS Support, but really looking forward for this feature or mean while if any alternate way we can achieve the desired results.

Kindly elaborate more on how to set flag driver-writes-secrets true?
[...]

PR #194, opened 2023-03-03
Adding Windows Support

saedx1:
*Issue #, if available:*
#131 

*Description of changes:*
- Closing a file before renaming it as this causes a problem on windows: `windows cannot access file because another process is using it`.
- In the translation of paths, unfortunately, `os.PathSeparator` isn't sufficient for windows. This is because windows also accepts `/` as a separator. Go's official `path_windows.go` actually does the same thing I'm adding https://go.dev/src/os/path_windows.go - see the first 16 lines of that file.
- I did build the image and push it to my personal github public container registry, so I can confirm that `Dockerfile.win` has been tested.
- I tested this on EKS with a windows node running 6 or more pods using secrets (as volumes and as env vars).

### DIFF
--- a/deployment/aws-provider-installer.yaml
+++ b/deployment/aws-provider-installer.yaml
@@ -88,3 +88,58 @@ spec:
             type: DirectoryOrCreate
       nodeSelector:
         kubernetes.io/os: linux
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  namespace: kube-system
+  name: csi-secrets-store-provider-aws-windows
+  labels:
+    app: csi-secrets-store-provider-aws-windows
+spec:
+  updateStrategy:
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      app: csi-secrets-store-provider-aws-windows
+  template:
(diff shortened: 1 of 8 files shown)


deployment/aws-provider-installer.yaml
```diff
@@ -88,3 +88,58 @@ spec:
...
+  template:
+    metadata:
+      labels:
+        app: csi-secrets-store-provider-aws-windows
+    spec:
+      serviceAccountName: csi-secrets-store-provider-aws
+      hostNetwork: false
+      containers:
+        - name: provider-aws-installer
+          image: ghcr.io/saedx1/secrets-store-csi-driver-provider-aws:1.0.0
```
saedx1, line 114: If you decide you will build and push the windows image (I hope so), this will need to change.

jbct:
Thank you for the PR. I'm going to close it at the moment, because we don't currently have plans to add Windows support. As an option, you may investigate using the `driver-writes-secrets` flag which may work around the issue you're encountering. Let's use #131 for discussion.

PR #221, opened 2023-05-01
Add rbac.serviceAccountName to helm chart values.yaml

mpriscella:
*Description of changes:*
Update to `values.yaml` to reflect that `rbac.serviceAccountName` is an available and configurable value in the current version (`0.3.2`) of the chart.


jbct:
Hi mpriscella, thank you for the pull request. Could you explain a bit more about your use case and the reason why you need serviceAccountName?

mpriscella:
Hi @jbct, I don't have a use case for this change, I just noticed that this value is already [referenced in _helpers.tpl](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/main/charts/secrets-store-csi-driver-provider-aws/templates/_helpers.tpl#L58) and didn't have a corresponding entry in `values.yaml`.

simonmarty (approved):

Issue #29, opened 2021-06-16
GRPC error: failed to mount secrets store object


Gearheads:
Hello:

We are currently experiencing an issue with the Secrets Store CSI Driver `v0.0.23`, where it is not able to mount the secret store object. We are seeing the Pods are stuck in a "ContainerCreating" state:
```
pod/robtest-aws-secret-manager-deploy-77747544cf-826n5   0/1     ContainerCreating   0          74m
pod/robtest-aws-secret-manager-deploy-77747544cf-8m9cd   0/1     ContainerCreating   0          74m
```

Here is the SecretProviderClass YAML:
```
apiVersion: secrets-store.csi.x-k8s.io/v1alpha1
kind: SecretProviderClass
metadata:
  name: robtest-aws-secret-provider-class
spec:
  provider: aws
  parameters:
    region: us-east-2
    objects: |
        - objectName: "arn:aws:secretsmanager:us-east-2:<AWS-account-#>:secret:MySecret-3jq4EL"
          objectType: "secretsmanager"
```

Any ideas as to why we are experiencing this issue below?

```
I0616 18:17:11.646404       1 nodeserver.go:300] "Using grpc client" provider="aws" 
  pod="robtest-aws-secret-manager-deploy-77747544cf-6pfvr"

I0616 18:19:11.643100       1 nodeserver.go:73] "unmounting target path as node publish volume failed" 
  targetPath="/var/lib/kubelet/pods/945bf1d1-a91b-4018-accd-1d79f5a1dbf9/volumes/kubernetes.io~csi/secrets-store-inline/mount"
  pod="cluster-addons/robtest-aws-secret-manager-deploy-77747544cf-6pfvr"

E0616 18:19:11.664341       1 utils.go:79] GRPC error: failed to mount secrets store objects for
  pod cluster-addons/robtest-aws-secret-manager-deploy-77747544cf-6pfvr,
  err: rpc error: code = Canceled desc = context canceled
```

Gearheads:
Turns out the issue was missing proxy values within the `csi-secrets-store-provider-aws` daemonset:
```
spec:
  serviceAccountName: csi-secrets-store-provider-aws
  containers:
    - name: provider-aws-installer
       imagePullPolicy: Always
       env:
       - name: HTTP_PROXY
         value: http://<proxy-url>:<proxy-port>
       - name: HTTPS_PROXY
         value: http://<proxy-url>:<proxy-port>
[...]

PR #227, opened 2023-05-22
Added env variables to helm chart daemonset

gavinelm:
*Issue #, if available:* https://github.com/aws/secrets-store-csi-driver-provider-aws/issues/29

*Description of changes:* Add an env field to the helm chart daemonset to implement the proxy settings solution described in https://github.com/aws/secrets-store-csi-driver-provider-aws/issues/29 to the helm chart. These changes were testing against the existing values.yaml and a values.yaml with the following added:
```
env:
  - name: HTTP_PROXY
    value: http://example.com:912
  - name: HTTPS_PROXY
    value: http://example.com:912
  - name: NO_PROXY
    value: 10.0.0.0/8
```


goyalya:
Thank you for submitting the request. AWS Secrets Manager is evaluating solutions to support proxy settings and would not be supporting it via copying environment variables. We recommend using VPC Endpoints as a potential workaround.

Issue #136, opened 2022-10-05
Secrets mounting limited by kubernetes client throttling


dindurthy:
**Describe the bug**
We routinely see errors like in the secrets store driver
```
"failed to mount secrets store object content\" err=\"rpc error: code = Unknown desc = rate: Wait(n=1) would exceed context deadline\" pod=\"<some-pod-name>\""
```
And in many cases we see corresponding errors in csi aws provider:
```
Failed to initialize AWS session\" err=\"rate: Wait(n=1) would exceed context deadline\"
```

The latter is coming from https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/c02c72deeebe81ae44cc00d9b931043f30455b82/server/server.go#L130.

The issue is I don't know what that error means. It looks to me like client-side throttling, but I'm unsure what the fundamental error is. I'm guessing that there is actual AWS server-side rate-limiting going on, but I'm not finding `ThrottlingException` or `Rate exceeded` or something definitive, and [it is notoriously a pain to work backwards from CloudTrail](https://aws.amazon.com/premiumsupport/knowledge-center/cloudtrail-rate-exceeded/).

This typically happens in the vicinity of upgrading services or draining nodes/scheduling onto new nodes, and it is node-specific. The node itself has free CPU and memory while these errors occur. 

**To Reproduce**

We tend to find this when draining a node of ~50 pods, each with 3-10 secret mounts.

Do you also notice this bug when using a different secrets store provider (Vault/Azure/GCP...)? **No**

If yes, the issue is likely with the k8s Secrets Store CSI driver, not the AWS provider. Open an issue in [that repo](https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues).

**Expected behavior**

`Failed to initialize AWS session` reports a meaningful error or some contingent error is separately logged.

**Environment:**
OS, Go version, etc.

**Additional context**

dindurthy:
Also when I look at the csi-driver-provider logs for `Servicing mount request for pod <some-pod-name>`, I don't see evidence of an exponential backoff
![Screen Shot 2022-10-05 at 4 46 22 PM](https://user-images.githubusercontent.com/14302694/194183608-6cc4d2bb-3cfa-43d7-957b-4b9d192dd691.png)
 Maybe that would be the secrets-store-driver's responsibility.

goyalya:
Thank you for reporting the issue. We are looking into the request.

dindurthy:
[...]

PR #323, opened 2024-03-13
Make throttling params QPS and Burst configurable

benjaminkz:
*Issue #, if available:*

*Description of changes:*

Kubernetes Go client uses a default value of 5 and 10 for QPS and burst. As a result, our customers may experience client-side rate limiting. This pull request makes them configurable to our customers.

Related issues: https://github.com/aws/secrets-store-csi-driver-provider-aws/issues/136


charts/secrets-store-csi-driver-provider-aws/templates/daemonset.yaml
```diff
@@ -31,6 +31,8 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - --provider-volume={{ .Values.providerVolume }}
+            - --qps={{ .Values.k8sThrottlingParams.qps | default 5 }}
```
simonmarty, line 34: Can we just not set these args if the value doesn't exist? This saves us from having to update the default value in multiple spots if it ever needs to be adjusted.

deployment/aws-provider-installer.yaml
```diff
@@ -62,6 +62,8 @@ spec:
           imagePullPolicy: Always
           args:
               - --provider-volume=/etc/kubernetes/secrets-store-csi-providers
+              - --qps=5
```
simonmarty, line 65: Same here, let's leave these out of the default YAML file. Customers can refer to the README for guidance on how to edit it to adjust these.

simonmarty (changes requested):

deployment/aws-provider-installer.yaml
benjaminkz: Will do.

README.md
```diff
@@ -220,6 +220,35 @@ envsubst < deployment/private-installer.yaml | kubectl apply -f -
 
 The AWS Secrets Manager and Config Provider provides compatibility for legacy applications that access secrets as mounted files in the pod. Security conscious applications should use the native AWS APIs to fetch secrets and optionally cache them in memory rather than storing them in the file system.
 
+### Client-Side Rate-Limitting to Kubernetes API server
+
+To mount each secret on each pod, the AWS CSI provider lookups the region of the pod and the role ARN associated with the service account by calling the K8s APIs. You can increase the value of qps and burst if you notice the provider is throttled by client-side limit to the API server.
```
adithyasolai, line 225: It's probably better to just write out "Kubernetes" instead of shortening to "K8s" to avoid confusion for readers.

README.md
```diff
@@ -220,6 +220,35 @@ envsubst < deployment/private-installer.yaml | kubectl apply -f -
```
adithyasolai, line 220: To keep our commit history clean, we prefer if you squash all your commits into a single commit before opening the PR.

In this scenario, the commit message of the single squashed commit would just be the title of this PR: "make throttling params qps and burst configurable"

README.md
simonmarty: Check with @ecraw-amzn

simonmarty (approved):

README.md
benjaminkz: Revised in new commit.

README.md
```diff
@@ -225,6 +225,12 @@ helm repo add aws-secrets-manager https://aws.github.io/secrets-store-csi-driver
 helm install -n kube-system secrets-provider-aws aws-secrets-manager/secrets-store-csi-driver-provider-aws --set useFipsEndpoint=true
 ```
 
+### Client-Side Rate-Limitting to Kubernetes API server
+
+To mount each secret on each pod, the AWS CSI provider lookups the region of the pod and the role ARN associated with the service account by calling the Kubernetes APIs. You can increase the value of qps and burst if you notice the provider is throttled by client-side limit to the API server.
+
+If you use Helm chart to install the provider, append the `--set-json 'k8sThrottlingParams={"qps": "<custom qps>", "burst": "<custom qps>"}'` flag in the install step.
```
ecraw-amzn, line 232: Is the flag for the CSI driver install or the Secrets Manager and Config Provider install?

benjaminkz: It's for the latter one. Our provider communicate use Kubernetes' Go client to communicate with the cluster, which may be throttled by these parameters.

ecraw-amzn (approved):

hannah-nguyen-ca (approved):

Issue #337, opened 2024-04-12
Allow setting `driver-writes-secrets` argument via Helm values


Elentary:
Providers binary supports specifying `driver-writes-secrets` bool argument to let the driver do the write instead of the plugin.
However, Helm chart currently does not allow to set this flag.

I see no reason in not having a Helm-chart value for this, especially taking into account the fact that this flag is required for deploying provider in self-hosted non-EKS clusters.

I would like to open a PR implementing this feature request if you do not mind.

jbct:
Hi @Elentary - thanks for the issue and corresponding PR. Do you have any details behind why you need this to be available in the helm chart?

PR #338, opened 2024-04-12
Allow setting driver-writes-secrets argument via Helm

Elentary:
Issue #337

*Description of changes:*

This change allows setting `driver-writes-secrets` flag via Helm chart value. One of usages example is setting the flag to `true` for deploying provider in non-EKS clusters in order to let the driver itself write secrets, because usual way of writing by plugin is not working.


simonmarty:
Closed in favor of #441, thanks for the contribution.

PR #384, opened 2024-07-23
Add Helm dependency on the Secrets Store CSI Driver

### LINKED ISSUE #345: Provider socket does not exist if provider pod starts before driver pod
**Describe the bug**
It appears that sometimes on our nodes the secrets-store-csi-driver-provider-aws pod will start before the secrets-store-csi-driver pod.  When that happens, our application pods that mount AWS secrets on such a node will get stuck in ContainerCreating status and the secrets-store-csi-driver pod generates logs such as:
```
"failed to mount secrets store object content" err="rpc error: code = Canceled desc = latest balancer error: connection error: desc = \"transport: Error while dialing: dial unix /etc/kubernetes/secrets-store-csi-providers/aws.sock: connect: no such file or directory\"" pod="mynamespace/mypod"
```
The secrets-store-csi-driver-provider-aws does not generate any logs which indicate an issue.

**To Reproduce**

Start the secrets-store-csi-driver-provider-aws pod will before the secrets-store-csi-driver pod and create an application pod that mounts AWS secrets. 

Do you also notice this bug when using a different secrets store provider (Vault/Azure/GCP...)? **Yes/No**
Haven't tried

If yes, the issue is likely with the k8s Secrets Store CSI driver, not the AWS provider. Open an issue in [that repo](https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues).

**Expected behavior**
The application pod mounts the secrets

**Environment:**
OS, Go version, etc.
EKS 1.29.1, secrets-store-csi-driver 1.4.2, secrets-store-csi-driver-provider-aws 1.0.r2-68-gab548b3-2024.03.20.21.58

**Additional context**
We noticed the secrets-store-csi-driver-provider-aws chart does not include `type: DirectoryOrCreate` in the `providervol` volume definition.  Is it possible when this hostpath doesn't exist on startup that the provider won't be able to use the socket but not generate any errors?

simonmarty:
*Issue #, if available:* Closes #345

*Description of changes:* Currently, the AWS provider is able to deploy before the Secrets Store CSI Driver. This can cause deployment failures since the volume mount path will not have been created yet. Introduce a dependency on version `1.5.2` of the Secret Store CSI Driver in the Helm chart.

Will probably pair this with a v2.0.0 release of the Helm chart.


dongheeJeong:
Hi @simonmarty. This MR looks great and seems ready to go! Any chance it could be merged soon when possible? Thanks for your work on this.

ThirdEyeSqueegee (approved):

ThirdEyeSqueegee (dismissed):

charts/secrets-store-csi-driver-provider-aws/values.yaml
```diff
@@ -46,3 +46,6 @@ securityContext:
 useFipsEndpoint: false
 
 dnsConfig: {}
+
+secrets-store-csi-driver:
+  install: true
```
simonmarty, line 51: We elected to make this default to true since we are doing a major version bump of the helm chart along with the Go SDK v2 release. This way we bundle backwards incompatible changes in the same release together while improving the customer experience by reducing the number of commands that need to be run to deploy the CSI driver.

We will update the README accordingly after the release.

ThirdEyeSqueegee (dismissed):

joebaro (approved):

ThirdEyeSqueegee (approved):

PR #394, opened 2024-09-12
Continue to fetch secret value even if describe access denied

suzaku:
*Issue #, if available:*

*Description of changes:*
Currently, if `DescribeSecret` is not allowed for one of the secrets, the entire `Mount` request and the auto rotation would fail. But sometimes secrets might be created with only `GetSecret` allowed, which is still enough permission to generate a new secret. It would be great to still try fetching secret value even if describe failed because of AccessDenied.

### DIFF
--- a/provider/secrets_manager_provider.go
+++ b/provider/secrets_manager_provider.go
@@ -2,8 +2,11 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"io/ioutil"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


simonmarty:
This is intentional design: `DescribeSecret` does not carry the decryption performance overhead that `GetSecretValue` does. (see the helper [here](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/b8df4953967406978a6199ab5321ee2308a387a7/provider/secrets_manager_provider.go#L160-L168)).

Grant `DescribeSecret` permissions to the identity fetching them in your cluster.

PR #52, opened 2021-09-30
LocalStack support

SizZiKe:
*Issue #, if available:*
#48 
*Description of changes:*
Adding the ability to override the sts, secretsmanager, and ssm endpoints separately (localstack feature), or to set a baseEndpoint if all services are hosted on the same URL.

I won't pretend to be a go expert, but the tests for server are already broken. I'd be happy if someone can guide me to write tests for this, but it's working ;)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #396, opened 2024-10-04
Service specific endpoints compatible resolver

sjones4:
This is an alternative to #52 allowing aws service endpoints to be configured as per the [service-specific endpoints](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html) feature defined environment variables which are [not otherwise supported](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-sdk-compat) in the aws sdk for go v1.

*Description of changes:*
An endpoint resolver is configured which delegates to the default resolver unless there is relevant configuration in the environment.

When environment configuration is found for service endpoints it is used unless disabled via `AWS_IGNORE_CONFIGURED_ENDPOINT_URLS=true`. The configuration can set all service endpoints (`AWS_ENDPOINT_URL`) or can configure [endpoints for specific services](https://docs.aws.amazon.com/sdkref/latest/guide/ss-endpoints-table.html).

This endpoint resolver would no longer be required after updating to the aws sdk for go v2.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
+	"github.com/aws/secrets-store-csi-driver-provider-aws/utils"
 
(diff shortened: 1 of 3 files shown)


sjones4:
This pr is no longer needed once #429 is merged.

Issue #397, opened 2024-10-08
Does not support fetching SSM parameters shared via AWS Resource Access Manager


thandleman-r7:
**Describe the bug**

Hey folks! 

We have a secret value we share with all AWS accounts via Resource Access Manager. I would like to leverage Secrets Store CSI Driver and this provider to mount the secret to a workload that relies on the value. However, this is currently not possible with the provider. The documentation states: `For SSM Parameter Store, this must be the Name of the parameter and can not be a full ARN.`

Because I am hard-headed, I tried anyway just to see what would happen. This results in an error which seems to occur because the name used to fetch the parameter does not match the name returned in the response. Stack trace include below.

This limitation makes it so we cannot use SSM Parameters shared via AWS RAM.

**Error On Mount**:
```
panic: runtime error: invalid memory address or nil pointer dereference                                                                                                                            │
│ [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16aac49]                                                                                                                            │
│                                                                                                                                                                                                    │
│ goroutine 20 [running]:                                                                                                                                                                            │
│ github.com/aws/secrets-store-csi-driver-provider-aws/provider.(*ParameterStoreProvider).fetchParameterStoreBatch(0xc0005916e0?, {0x0, {0xc000700050, 0xe}, {0x20015a0, 0xc000610508}}, {0x1fc2108, │
│     /Users/martysi/Documents/workspaces/upstream_repos/secrets-store-csi-driver-provider-aws/provider/parameter_store_provider.go:157 +0x769                                                       │
│ github.com/aws/secrets-store-csi-driver-provider-aws/provider.(*ParameterStoreProvider).fetchParameterStoreValue(0xc000591620, {0x1fc2108, 0xc000144480}, {0xc0006104f0, 0x1, 0x1}, 0x16acd10?)    │
│     /Users/martysi/Documents/workspaces/upstream_repos/secrets-store-csi-driver-provider-aws/provider/parameter_store_provider.go:85 +0x1b1                                                        │
│ github.com/aws/secrets-store-csi-driver-provider-aws/provider.(*ParameterStoreProvider).GetSecretValues(0x1837d80?, {0x1fc2108, 0xc000144480}, {0xc0006104f0, 0x1, 0x1}, 0xc00059a900?)            │
[...]

PR #423, opened 2025-01-24
Support passing the ARN for an SSM Parameter via objectName

thandleman-r7:
*Issue #, if available:* https://github.com/aws/secrets-store-csi-driver-provider-aws/issues/397

*Description of changes:*

The documentation states that you cannot pass an ARN for an object type of `ssmparameter`. This limitation seems to only be imposed by the fact that, after fetching the parameters from SSM, we only check if the parameter name matches one of the descriptors. However, the ARN is also available in the response for each parameter. As can be seen in the linked issue, the call to SSM doesn't fail, it is matching the result to the descriptor that fails due to a nil pointer.

This change falls back on the ARN if a descriptor is not matched by name. We can assume that if the parameter in the results was not found by name, then the ARN was passed.

### DIFF
--- a/provider/parameter_store_provider.go
+++ b/provider/parameter_store_provider.go
@@ -150,7 +150,13 @@ func (p *ParameterStoreProvider) fetchParameterStoreBatch(
 	// Build up the results from the batch
 	for _, parm := range rsp.Parameters {
 
-		descriptor := batchDesc[*(parm.Name)]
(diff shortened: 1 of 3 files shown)


ltflb-bgdi:
@simonmarty is there any chance that this PR can be merged in near future? This issue is blocking us from properly implementing ssm support in our environment. Thanks in advance for your support.

PR #431, opened 2025-02-11
Remove deprecated ioutil calls

micahhausler:
*Issue #, if available:*

N/A

*Description of changes:*

Removes deprecated ioutil calls

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -297,7 +296,7 @@ func validateMounts(t *testing.T, dir string, tst testCase, rsp *v1alpha1.MountR
 
 	// Check for the expected secrets
 	for file, val := range tst.expSecrets {
-		secretVal, err := ioutil.ReadFile(filepath.Join(dir, file))
+		secretVal, err := os.ReadFile(filepath.Join(dir, file))
 		if err != nil {
 			t.Errorf("%s: Can not read file %s", tst.testName, file)
 			return false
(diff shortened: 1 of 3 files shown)


simonmarty (approved):

micahhausler:
Last push was just a rebase

server/server_test.go
```diff
@@ -1994,11 +1993,7 @@ func TestMounts(t *testing.T) {
 
 		t.Run(tst.testName, func(t *testing.T) {
 
-			dir, err := ioutil.TempDir("", strings.Map(nameMapper, tst.testName))
-			if err != nil {
-				panic(err)
-			}
-			defer os.RemoveAll(dir) // Cleanup
+			dir := t.TempDir() // t.TempDir() handles cleanup automatically
```
simonmarty, line 1996: Previously we would create temporary directories, in this case in the pattern `strings.Map(nameMapper, tst.testName)`. I wonder if we should be worried about potential collisions between different tests' use of filesystem paths. Maybe it's best to stick to the current approach, unless you happen to know better.

I see `os` has a similar [CreateTemp](https://pkg.go.dev/os#CreateTemp) function.

simonmarty (changes requested):

server/server_test.go
micahhausler: I don't think we have to worry about this based on the `t.TempDir()` implementation. (You can click the [method name](https://pkg.go.dev/testing#T.TempDir) to see the source code.) It uses the test name as the unique path.

https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/testing/testing.go;l=1224-1271

vedant-jaiswal:
Handled in this PR(https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/463)

PR #433, opened 2025-02-11
Use Go conventions on context.Context

micahhausler:
*Issue #, if available:*

N/A

*Description of changes:*

Go has strong convention around passing context.Context as the first call to functions. This is called out in the [context package documentation](https://pkg.go.dev/context)

> The Context should be the first parameter, typically named ctx:
> ```
> func DoSomething(ctx context.Context, arg Arg) error {
>	// ... use ctx ...
> }
> ```
> Do not pass a nil [Context](https://pkg.go.dev/context#Context), even if a function permits it. Pass [context.TODO](https://pkg.go.dev/context#TODO) if you are unsure about which Context to use.

  
I intentionally did not fix `auth/` and `credential_provider/` misuses of context, as that will be fixed in a later PR updating the AWS Go SDK.

### DIFF
--- a/provider/parameter_store_provider.go
+++ b/provider/parameter_store_provider.go
@@ -79,7 +79,7 @@ func (p *ParameterStoreProvider) fetchParameterStoreValue(
 ) (values []*SecretValue, err error) {
 
 	for _, client := range p.clients {
-		batchValues, err := p.fetchParameterStoreBatch(client, ctx, batchDescriptors, curMap)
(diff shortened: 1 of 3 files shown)


simonmarty (approved):

vedant-jaiswal:
Handled in this PR(https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/463)

Issue #263, opened 2023-08-23
Capability to set file-system permissions for mounted secrets


milokrisko:
**Describe the solution you'd like**

Supposing my secret is an ssh key. Then I'm going to want the file to have permissions 0600, else I'll get

`@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@`

**Anything else you would like to add:**

Common advice seems to be to run an initContainer and adjust the permissions on files as needed from that. But the secrets-store-csi-driver insists that I mount the volume as read-only, so that doesn't work.

This request was copied from the https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/722 
As the resolution from there was to submit this as Request under specific cloud provider and i didn't notice any request to be filled for AWS as for now.

milokrisko:
My Current workaround is "terrible".

I forced the secret-store-csi-driver to sync the files as k8s secret (which i didn't need before) and then i set up to mount this secret as volume where those change are enabled via k8s:

`        - name: ssh_keys
          secret:
            defaultMode: 420
            items:
            - key: users.json
              path: users.json
            - key: ssh_host_ed25519_key
              mode: 256
              path: ssh_host_ed25519_key
            - key: ssh_host_rsa_key
              mode: 256
              path: ssh_host_rsa_key
            secretName: {{ .Release.Name  }}-secret-fs`
            
This way the mounting works even though i do have this configuration there twice under diff paths for example:
/data/keys/aws
and from k8s it's synced into
/data/keys/

jbct:
Thank you for the feature enhancement request. We will look into this.

colearendt:
We are encountering the same problem. We have a product that requires 600 permissions on its secrets.

Just for clarity, we used the same workaround / solution discussed above:
- create a `SecretObject` in the `SecretProviderClass`
- mount a secret as an env var (to ensure that the kubernetes secret gets created)
- mount the secret to disk and set `defaultMode`

Tedious to set up and clutters our kubernetes cluster with an extra secret resource. Worth noting that permissions are also pretty complicated in Kubernetes Secrets / volume mounts.

joebowbeer:
[...]

PR #435, opened 2025-02-20
Add File Permission Support

liaandy:
*Issue #, if available:* https://github.com/aws/secrets-store-csi-driver-provider-aws/issues/263

*Description of changes:*
Introduce an optional parameter filePermission to support configuration of file permissions.


server/server.go
```diff
@@ -79,6 +79,12 @@ func NewServer(
 // Store or Secrets Manager) and write the secrets to the mount point. The
 // version ids of the secrets are then returned to the driver.
 func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.MountRequest) (response *v1alpha1.MountResponse, e error) {
+	// Log out the write mode
+	if s.driverWriteSecrets == true {
```
simonmarty, line 83: It's been a while since I've written go, can't you just
```go
if s.driverWriteSecrets {
...
}
```
I don't think that evaluation is necessary.

server/server_test.go
```diff
@@ -287,6 +287,62 @@ func buildMountReq(dir string, tst testCase, curState []*v1alpha1.ObjectVersion)
 
 }
 
+func resolveFilePath(fileName string, translate string) string {
```
simonmarty, line 290: Can you add a docstring explaining what this method does/why it's needed?

simonmarty (changes requested):

provider/secret_descriptor.go
```diff
@@ -181,6 +193,24 @@ func (p *SecretDescriptor) GetObjectVersion(useFailoverRegion bool) (secretName
...
+// This function validates the filePermission and ensures it is a valid 4 digit octal string
+func (p *SecretDescriptor) validateFilePermission(filePermission string) error {
+	// No file permission
+	if len(filePermission) == 0 {
+		return nil
+	}
+
+	match, _ := regexp.MatchString("^[0-7]{4}$", filePermission)
+
+	if match == false {
```
simonmarty, line 207: `if !match {`

simonmarty (changes requested):

README.md
```diff
@@ -192,6 +192,7 @@ The primary objects field of the SecretProviderClass can contain the following s
 * objectName: This field is required. It specifies the name of the secret or parameter to be fetched. For Secrets Manager this is the [SecretId](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#API_GetSecretValue_RequestParameters) parameter and can be either the friendly name or full ARN of the secret. For SSM Parameter Store, this is the [Name](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html#API_GetParameter_RequestParameters) of the parameter and can be either the name or full ARN of the parameter.
 * objectType: This field is optional when using a Secrets Manager ARN for objectName, otherwise it is required. This field can be either "secretsmanager" or "ssmparameter".
 * objectAlias: This optional field specifies the file name under which the secret will be mounted. When not specified the file name defaults to objectName.
+* filePermission: This optional field expects a 4 digit string which specifies the file permission for the secret that will be mounted. When not specified this will default to "0644" permisions. Ensure the 4 digit string is a valid octal file permission.
```
joebaro, line 195: > When not specified this will default to "0644" permisions
It will probably be more accurate to say something like "When not specified this will default to the permissions passed from the CSI driver (0644)"

Also there is a typo in "permisions".

provider/secret_descriptor.go
```diff
@@ -181,6 +193,24 @@ func (p *SecretDescriptor) GetObjectVersion(useFailoverRegion bool) (secretName
...
+// Private helper to validate a filePermission
+//
+// This function validates the filePermission and ensures it is a valid 4 digit octal string
+func (p *SecretDescriptor) validateFilePermission(filePermission string) error {
+	// No file permission
+	if len(filePermission) == 0 {
+		return nil
+	}
+
+	match, _ := regexp.MatchString("^[0-7]{4}$", filePermission)
```
joebaro, line 205: For performance reasons, when using a fixed RE we prefer to pre-compile the expression. See `regexp.MustCompile()` above.

provider/secret_descriptor.go
```diff
@@ -145,11 +151,17 @@ func (p *SecretDescriptor) GetSecretType() (stype SecretType) {
 
 // Return a descriptor for a jmes object entry within the secret
 func (p *SecretDescriptor) getJmesEntrySecretDescriptor(j *JMESPathEntry) (d SecretDescriptor) {
+	permission := j.FilePermission
```
joebaro, line 154: We are not also validating the permissions string here.

server/server.go
```diff
@@ -168,9 +174,14 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
...
 	var files []*v1alpha1.File
+	var permission os.FileMode
 	for _, secret := range fetchedSecrets {
-
-		file, err := s.writeFile(secret, filePermission)
+		permission = filePermission
+		if secret.Descriptor.FilePermission != "" {
+			parsedPermission, _ := strconv.ParseInt(secret.Descriptor.FilePermission, 8, 32)
+			permission = os.FileMode(parsedPermission)
+		}
```
joebaro, line 183: It would be better to not embed the logic of how the permissions are handled in the server code. Instead what we could do is create a `getFilePermissions()` method on SecretDescriptor that returns either the default permission or the override permission. The default permission could be passed into the SecretDescriptor in `NewSecretDescriptorList()` above the same way we pass in the region.

provider/secret_descriptor.go
```diff
@@ -145,11 +163,17 @@ func (p *SecretDescriptor) GetSecretType() (stype SecretType) {
...
+	if permission == "" {
```
simonmarty, line 167: Nit: This guard clause is different from the ones on L210 and L222. Is there a reason for that? No need to fix, just making sure I'm not missing anything.

provider/secret_descriptor.go
```diff
@@ -14,6 +15,12 @@ import (
 // An RE pattern to check for bad paths
 var badPathRE = regexp.MustCompile("(/\\.\\./)|(^\\.\\./)|(/\\.\\.$)")
 
+// An RE pattern to check for valid file permission
+var validFilePermissionRE = regexp.MustCompile("^[0-7]{4}$")
```
simonmarty, line 19: Should we perform validation to ensure that the passed perms actually allow the driver to write the secret or do we just let it fail? https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/5f41c0004924d6e322eccd38432eb8a936a9b44a/server/server.go#L318-L326

Say they set `0444` as the file permission, the write will fail, should we pre-empt this failure in the validation?

simonmarty (changes requested):

provider/secret_descriptor.go
```diff
@@ -14,6 +15,12 @@ import (
...
+
+// Default file permission
+var defaultFilePermission = os.FileMode(0644)
```
joebaro, line 22: Do not hard code the default permissions. We should be passing in the permissions [passed in from the driver](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/main/server/server.go#L115) as the default. They can be passed into [NewSecretDescriptorList](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/main/server/server.go#L150) and then just call `SetDefaultFilePermission()` on each entry.

joebaro (approved):

simonmarty (approved):

mtjakobczyk:
Thank you for merging this new feature. Do you already have in mind the release date?

Issue #437, opened 2025-02-26
Find a way to obtain the AWS region without listing pods and nodes


simonmarty:
For [listing Pods and Nodes](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/be94dd016af342d6e42ce5e1c186bee6b40ce3fb/server/server.go#L255-L262), it looks like thats only done for getting the AWS region. It'd be great if this could be provided by a Helm Chart environment variable in the short term. Longer term, if/when https://github.com/kubernetes/kubernetes/issues/40610, that could just be injected into the pod without any user input.

Its worth noting that Kubelet today has permissions to list all pods (but not nodes), but this permission [is going to be restricted](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/4601-authorize-with-selectors/README.md) in future versions of Kubernetes to only allow a node to list the pods running on it.

ThirdEyeSqueegee:
Closed by #445

Issue #422, opened 2025-01-23
Remove RBAC permission from ASM plugin


micahhausler:
The ASM plugin currently has four Kubernetes API permissions that are all cluster-scoped (not node-scoped), and could benefit from paring them down by making certain ones optional or removing them altogether. Ideally, the ASM plugin would not require any permissions of its own from Kubernetes or AWS, so that the principal of least permissions is followed.

From https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/main/deployment/private-installer.yaml#L13-L24
```yaml
- apiGroups: [""]
  resources: ["serviceaccounts/token"]
  verbs: ["create"]
- apiGroups: [""]
  resources: ["serviceaccounts"]
  verbs: ["get"]
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["get"]
- apiGroups: [""]
  resources: ["nodes"]
  verbs: ["get"]
```

* #400 references how ASM could leverage TokenRequest to get tokens via the Kubelet
* `GET` on `serviceaccounts` is used for IRSA-based auth to get [a pod's IAM role via an annotation](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/be94dd016af342d6e42ce5e1c186bee6b40ce3fb/auth/auth.go#L113). For installations using EKS Pod Identity, this can be removed (you could use a condition in the helm chart template to make this optional)
* For [listing Pods and Nodes](https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/be94dd016af342d6e42ce5e1c186bee6b40ce3fb/server/server.go#L255-L262), it looks like thats only done for getting the AWS region. It'd be _great_ if this could be provided by a Helm Chart environment variable in the short term. Longer term, if/when [Kubernetes adds support for node labels to the Downward API](https://github.com/kubernetes/kubernetes/issues/40610), that could just be injected into the pod without any user input. 
  Its worth noting that Kubelet today has permissions to list all pods (but not nodes), but this permission [is going to be restricted](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/4601-authorize-with-selectors/README.md) in future versions of Kubernetes to only allow a node to list the pods running on it.

PR #445, opened 2025-03-13
Remove pod and node RBAC permissions using AWS_REGION env var

vedant-jaiswal:
*Issue #437*

*Description of changes:*

This PR partially addresses #422 by implementing #437 - removing pod and node RBAC permissions by making region discovery configurable via environment variable.

Changes made:
1. Added `awsRegion` field to values.yaml for configuring AWS region
2. Made pod/node permissions conditional in rbac.yaml based on awsRegion setting
3. Added AWS_REGION environment variable to container in daemonset.yaml
4. Updated server.go to check AWS_REGION env var before falling back to pod/node lookup
5. Added tests for region retrieval logic

Testing:
- Added unit tests for region retrieval logic in server_test.go
- Manually tested with and without AWS_REGION set:
  - With awsRegion: pod/node permissions removed, AWS_REGION env var set
  - Without awsRegion: falls back to pod/node permissions for region discovery
- Verified RBAC permissions are properly conditional


ThirdEyeSqueegee (approved):

server/server_test.go
```diff
@@ -2579,6 +2579,55 @@ func TestNoPath(t *testing.T) {
...
+	region, err := svr.getRegionFromNode(context.TODO(), "fakeNS", "fakePod")
+
+	if err != nil {
+		t.Fatalf("Expected no error with AWS_REGION set, got: %v", err)
+	}
+	if region != "us-west-2" {
+		t.Fatalf("Expected region us-west-2, got: %s", region)
+	}
+
+	// Test fallback to node labels when AWS_REGION not set
```
JoeJesse, line 2605: In its current state this single test is exercising three cases. I think it would be better to separate it out such that there is one test for each case.

vedant-jaiswal: I've addressed your comment by splitting the single test into three separate test cases:

1. TestGetRegionFromNodeWithAWSRegionEnvVar: Tests when AWS_REGION is set
2. TestGetRegionFromNodeWithNodeLabels: Tests fallback to node labels when AWS_REGION is not set
3. TestGetRegionFromNodeError: Tests the error case when no region is available

ThirdEyeSqueegee (approved):

Issue #95, opened 2022-05-27
Can't reference AWSCURRENT and AWSPREVIOUS in the same secretproviderclass


sergei-riazanov-privitar:
The configuration I want to use:
```
spec:
  parameters:
    objects: |
      ---
      - objectAlias: "encryption-key"
        objectVersionLabel: "AWSCURRENT"
        objectName: "arn:aws:secretsmanager:XXX"
        objectType: "secretsmanager"
      - objectAlias: "previous-encryption-key"
        objectVersionLabel: "AWSPREVIOUS"
        objectName: "arn:aws:secretsmanager:XXX"
        objectType: "secretsmanager"
  provider: aws
  secretObjects:
  - data:
    - key: encryption-key
      objectName: encryption-key
    - key: previous-encryption-key
      objectName: previous-encryption-key
    secretName: YYY
    type: Opaque
```

As result I get this on mounting the volume:
"Warning  FailedMount             71s (x9 over 3m19s)  kubelet                     MountVolume.SetUp failed for volume "YYY" : rpc error: code = Unknown desc = failed to mount secrets store objects for pod ..., err: rpc error: code = Unknown desc = Name already in use for objectName: arn:aws:secretsmanager:XXX"

sergei-riazanov-privitar:
Does [https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/c02c72deeebe81ae44cc00d9b931043f30455b82/provider/secret_descriptor.go#L262](this check) make sense if the object have aliases?

jbct:
Thank you for the report, sergei-riazanov-, we are looking into the issue.

dnlopes:
Just a quick question on the behavior after this change:

by putting together both `AWSPREVIOUS` and `AWSCURRENT`, does this mean that both secrets are refreshed at the same time?

PR #446, opened 2025-03-17
chore: update/cleanup workflows

ThirdEyeSqueegee:
*Issue #, if available:*
N/A

*Description of changes:*

Closes #407, #406, #405

- Add `types` rule for pull requests
- Bump action versions
- Fix formatting

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #447, opened 2025-03-18
fix: make secret descriptor validation more granular (#95)

ThirdEyeSqueegee:
*Issue #, if available:*
#95 

*Description of changes:*
- Make secret descriptor validation more granular to account for edge cases as in #95 
- Add tests for various combinations of conflicting `objectName`s, `objectAlias`es, and `objectVersionLabel`s 

### DIFF
--- a/provider/secret_descriptor.go
+++ b/provider/secret_descriptor.go
@@ -56,6 +56,13 @@ type SecretDescriptor struct {
 	mountDir string `json:"-"`
 }
 
+// Slice of the above type used for validation in NewSecretDescriptorList
+type SecretDescriptorSlice struct {
+	ObjectAlias string `json:"objectAlias"`
+
+	ObjectVersionLabel string `json:"objectVersionLabel"`
+}
+
 // An individual json key value pair to mount
 type JMESPathEntry struct {
 	//JMES path to use for retrieval
@@ -385,7 +392,8 @@ func NewSecretDescriptorList(mountDir, translate, objectSpec string, regions []s
 
 	// Validate each record and check for duplicates
 	groups := make(map[SecretType][]*SecretDescriptor, 0)
-	names := make(map[string]bool)
+	seenNames := make(map[string]SecretDescriptorSlice)
+	seenAliases := make(map[string]bool)
 	for _, descriptor := range descriptors {
 
 		descriptor.translate = translate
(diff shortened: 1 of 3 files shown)


provider/secret_descriptor.go
```diff
@@ -337,29 +345,70 @@ func NewSecretDescriptorList(mountDir, translate, objectSpec string, regions []s
...
+		// -------------------------------------------
+		// | # | OBJECT ALIAS | OBJECT VERSION LABEL |
+		// |---|--------------|----------------------|
+		// | 1 | duplicate    | empty                |
+		// | 2 | empty        | duplicate            |
+		// | 3 | duplicate    | duplicate            |
+		// | 4 | empty        | empty                |
+		// -------------------------------------------
+		found, ok := seenNames[descriptor.ObjectName]
+		if ok {
```
simonmarty, line 359: Use guard causes to remove nesting? (exit early)
```
if !ok {
// exit block
}

if !descHasAlias || !foundHasAlias {
// exit
}

...
```

provider/secret_descriptor.go
```diff
@@ -337,29 +345,70 @@ func NewSecretDescriptorList(mountDir, translate, objectSpec string, regions []s
...
 
 		if len(descriptor.JMESPath) == 0 { //jmesPath not used. No more checks
 			continue
 		}
 
 		for _, jmesPathEntry := range descriptor.JMESPath {
-			if names[jmesPathEntry.ObjectAlias] {
-				return nil, fmt.Errorf("Name already in use for objectAlias: %s", jmesPathEntry.ObjectAlias)
+			if seenAliases[jmesPathEntry.ObjectAlias] {
+				return nil, fmt.Errorf("found duplicate object alias %s in JMES path entry %s", jmesPathEntry.ObjectAlias, jmesPathEntry.Path)
```
simonmarty, line 409: I'm a bit concerned about this breaking customers that search for this error message, should we keep the start of the string the same?

simonmarty (changes requested):

simonmarty:
Could you also update the PR title to something more descriptive of the changes we're making?

provider/secret_descriptor.go
ThirdEyeSqueegee: Inverting the `if`s would honestly be messier since there's `else` cases for almost everything too and we want to run the rest of the validation logic after the `if ok` block; I think we should leave it as is unless there's a strong reason to invert (personally I don't think there's enough levels of nesting to necessitate inverting)

dnlopes:
I'm _very_ interested in having this PR merged and added to the next release, for the same reasons the original issue was reported.

So, let me know if anything is needed, I can try to help.

dnlopes:
Just a quick question on the behavior after this change:

by putting together both `AWSPREVIOUS` and `AWSCURRENT`, does this mean that both secrets are refreshed at the same time?

provider/secret_descriptor.go
simonmarty: Something like

```
	if ok {
			descHasAlias := descriptor.ObjectAlias != ""
			foundHasAlias := found.ObjectAlias != ""
			descHasVersionLabel := descriptor.ObjectVersionLabel != ""
			foundHasVersionLabel := found.ObjectVersionLabel != ""

			errorPrefix := fmt.Errorf("found descriptor with duplicate object name %s", descriptor.ObjectName)

			if descHasAlias && foundHasAlias && descriptor.ObjectAlias == found.ObjectAlias {
				if !descHasVersionLabel || !foundHasVersionLabel {
					// Case 1
					return nil, fmt.Errorf("%s, object alias %s with no version label", errorPrefix, descriptor.ObjectAlias)
				}
				
				if descriptor.ObjectVersionLabel == found.ObjectVersionLabel {
						// Case 3
					return nil, fmt.Errorf("%s, object alias %s, and version label %s", errorPrefix, descriptor.ObjectAlias, descriptor.ObjectVersionLabel)
				}
			} else {
				if !descHasVersionLabel || !foundHasVersionLabel {
					// Case 4
					return nil, fmt.Errorf("%s with no object alias or version label", errorPrefix)
				}

				if descriptor.ObjectVersionLabel == found.ObjectVersionLabel {
					// Case 2
					return nil, fmt.Errorf("%s and version label %s with no object alias", errorPrefix, descriptor.ObjectVersionLabel)
				}
			}
		}
```

There might be a way to denest this one step further even, it's a matter of finding which cases cause a return an short circuit those as early as possible in the code

It makes your code easier to test because you can clearly tell all the failure modes since they become sequential and they come up early in the subroutine

simonmarty: Also, nit: maybe `ok` can be renamed to something more descriptive?

simonmarty (changes requested):

provider/secret_descriptor.go
ThirdEyeSqueegee: Thanks, I totally blanked on the else-after-returns. And yeah I can rename `ok`, only reason I used it is that it appears to be the conventional name for the second variable returned during lookups

simonmarty: If it's idiomatic you can keep it.

provider/secret_descriptor_test.go
```diff
@@ -150,13 +151,148 @@ func TestConflictingName(t *testing.T) {
...
+          objectType: ssmparameter`
+
+	_, err := NewSecretDescriptorList("/", "", objects, singleRegion)
+
+	if err != nil {
+		t.Fatalf("Expected no error, got error: %v", err)
+	}
+}
+
+// Conflicting name; version label and alias present -- should not throw
```
simonmarty, line 276: Could you update the test comments to help identify which "Cases" above each one of these matches to? That's gonna make future referencing easier.

ThirdEyeSqueegee: Good call, will do

provider/secret_descriptor_test.go
```diff
@@ -584,7 +720,7 @@ func TestVersionIdsMustMatch(t *testing.T) {
     - objectName: "MySecret1"
       objectType: ssmparameter
       objectVersion:  OldVersionId
-      failoverObject: 
+      failoverObject:
```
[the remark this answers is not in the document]
ThirdEyeSqueegee: Sure thing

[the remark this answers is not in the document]
ThirdEyeSqueegee: There was already a workflow step to check formatting but I don't think it was working, swapped it out for an action that checks formatting using `gofmt` in #446

dnlopes:
Hello all @simonmarty @ThirdEyeSqueegee, first of all thanks for this PR, it will address a blocking issue with one of our use cases.

What can be done to move this PR forward? I'm willing to help if needed, but it seems everything is mostly done.

What is missing? Just bandwidth from the codeowners to review this?

Thanks in advance!

provider/secret_descriptor_test.go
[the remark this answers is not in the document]
simonmarty: Spoke offline, done in https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/446

simonmarty (approved):

gnaikrah (approved):

PR #458, opened 2025-05-04
fix: change http timeout for pod identity from 100ms to 10s

Mmx233:
*Issue #, if available:*

*Description of changes:*

The 100ms hard-coded timeout set for the pod identity HTTP client is too short and may cause issues with volume mounting in certain scenarios, especially in regions without the eks-auth VPCE. When the driver encounters a cancel error, it often requires complex debugging to identify the specific cause, which is not user-friendly. This PR adjusts the timeout to 10s to prevent such issues.

### DIFF
--- a/credential_provider/pod_identity_credential_provider.go
+++ b/credential_provider/pod_identity_credential_provider.go
@@ -20,7 +20,7 @@ const (
 	podIdentityAudience   = "pods.eks.amazonaws.com"
 	defaultIPv4Endpoint   = "http://169.254.170.23/v1/credentials"
 	defaultIPv6Endpoint   = "http://[fd00:ec2::23]/v1/credentials"
-	httpTimeout           = 100 * time.Millisecond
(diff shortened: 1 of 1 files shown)


simonmarty:
Thanks for the contribution. How did you land on 10s?

Mmx233:
10 seconds is an appropriate value that ensures requests can complete even under extreme conditions, such as potentially slow redirects. At the same time, if something does go wrong, an error will be shown after ten seconds, which is not too long to wait.

To be honest, the best approach is still to allow users to configure it themselves, such as by reading from an environment variable.

simonmarty:
> To be honest, the best approach is still to allow users to configure it themselves, such as by reading from an environment variable.

Agreed.

Hang tight on this one, I'm digging for information on how we landed on 100ms in the first place.

Hymian7:
@simonmarty is there a timeline for this PR or a decision whether it will be included or not?

PR #466, opened 2025-06-02
feat: add support for base64 encoded secrets fields

mickaelicoptere:
*Description of changes: 
add support for base64 encoded secret fields

### DIFF
--- a/provider/secret_descriptor.go
+++ b/provider/secret_descriptor.go
@@ -40,6 +40,9 @@ type SecretDescriptor struct {
 	// One of secretsmanager or ssmparameter (not required when using full secrets manager ARN).
 	ObjectType string `json:"objectType"`
 
+	// Optional encoding type of the secret value (e.g., "base64")
+	ObjectEncoding string `json:"objectEncoding"`
+
 	// Optional file permission (default to driver file permission).
 	FilePermission string `json:"filePermission"`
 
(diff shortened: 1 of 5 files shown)


simonmarty:
What's the use case for this?

mickaelicoptere:
> What's the use case for this?

@simonmarty 
Some fields in our secret's JSON are base64 encoded because they may contain data that can't be represented as UTF-8 (like binary content). By handling the decoding directly in the secrets-store-csi provider, we eliminate the need for a separate initContainer to perform this transformation.

simonmarty:
You shouldn't have to base64 encode binary into AWS Secrets Manager, ASM supports binary secret values
https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#SecretsManager-GetSecretValue-response-SecretBinary

We already support retrieving the secret binary here https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/529f812842b350baf1ae599fd7a01cda01415254/provider/secrets_manager_provider.go#L236-L242

mickaelicoptere:
What I intended was to store these base64-encoded values within a secret string formatted like this:
```json
{
"key1": "value1",
"key2": "dmFsdWUyCg=="
}
```
This approach would be somewhat similar to how it's done in Azure's secret-store-csi-provider: https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/examples/keyvault-certs/v1alpha1_secretproviderclass_certs.yaml#L40

MXEQ-oddball:
@simonmarty It's unfortunate that this PR got closed unmerged. This would be a great feature and would improve the flexibility of this CSI Driver Provider. Any chance you would reconsider?

PR #472, opened 2025-06-13
feat: make `httpTimeout` configurable

ThirdEyeSqueegee:
*Issue #, if available:*
#458 

*Description of changes:*
Add `httpTimeout` config option 


auth/auth.go
```diff
@@ -80,9 +84,9 @@ func (p Auth) GetAWSConfig(ctx context.Context) (aws.Config, error) {
 	var credProvider credential_provider.ConfigProvider
 
 	if p.usePodIdentity {
-		klog.Infof("Using Pod Identity for authentication in namespace: %s, service account: %s", p.nameSpace, p.svcAcc)
+		klog.Infof("Using Pod Identity for authentication in namespace: %s, service account: %s, httpTimeout: %v", p.nameSpace, p.svcAcc, p.httpTimeout)
```
simonmarty: Should we error out if httpTimeout is set but the user is running IRSA?

server/server.go
```diff
@@ -107,12 +109,32 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
...
+			return nil, fmt.Errorf("failed to parse httpTimeout value '%s': %v", httpTimeoutStr, err)
+		}
+		if httpTimeoutDuration <= 0 {
+			return nil, fmt.Errorf("httpTimeout must be positive, got: %v", httpTimeoutDuration)
+		}
+		if httpTimeoutDuration > 30*time.Second {
+			klog.Warningf("httpTimeout value %v is unusually high, consider using a smaller value", httpTimeoutDuration)
+		}
+	} else {
+		// Default to 100ms for backward compatibility
```
simonmarty: I'm nitpicking, but we default to 100ms. It's not (just) for backwards compatibility, it's because we should have a default so customers don't have to configure this if they don't have a need to.

Probably just trim the comment down to `Default to 100ms` :)

server/server_test.go
```diff
@@ -258,6 +258,11 @@ func buildMountReq(t *testing.T, dir string, tst testCase, curState []*v1alpha1.
 		attrMap["usePodIdentity"] = usePodIdentity
 	}
 
+	httpTimeout := tst.attributes["httpTimeout"]
+	if len(httpTimeout) > 0 {
```
simonmarty: What's this check for?

auth/auth.go
ThirdEyeSqueegee: Why? Isn't just ignoring it sufficient in that case?

server/server_test.go
ThirdEyeSqueegee: It's just the same pattern as the assignments above to set the key in the map if `httpTimeout` is non-empty

README.md
```diff
@@ -187,6 +187,7 @@ The parameters section contains the details of the mount request and contain one
   - "ipv4", "IPv4", or "IPV4" - Force the use of Pod Identity Agent IPv4 endpoint
   - "ipv6", "IPv6", or "IPV6" - Force the use of Pod Identity Agent IPv6 endpoint
   - not specified - Use auto endpoint selection, trying IPv4 endpoint first and falling back to IPv6 endpoint if IPv4 fails
+* httpTimeout: An optional field that specifies the HTTP timeout duration for Pod Identity Agent communication. This field is only applicable when using EKS Pod Identity feature and will be ignored when using IAM Roles for Service Accounts. The value must be a valid Go duration string (e.g., "100ms", "1s", "500ms"). If not specified, defaults to "100ms". Values are case-sensitive and must be positive. Example values: "50ms", "200ms", "1s", "2s".
```
simonmarty: You've added example values twice to this line.

auth/auth.go
simonmarty: To avoid someone configuring a different timeout on IRSA expecting a behavior change. 

An acceptable alternative would be to better qualify the parameter. `podIdentityHttpTimeout` or something of the sort.

simonmarty:
I do want to revisit whether or not we should allow configuring this value at the driver level (similar to the discussion we had about permissions configuration).

auth/auth_test.go
```diff
@@ -38,18 +39,122 @@ var sessionTests []sessionTest = []sessionTest{
...
+	tests := []struct {
+		name                   string
+		region                 string
+		nameSpace              string
+		svcAcc                 string
+		podName                string
+		preferredAddressType   string
+		usePodIdentity         bool
+		podIdentityHttpTimeout time.Duration
+		expectError            bool
```
i-am-SR, line 52: Is this supposed to always be false? If, so can we remove this? Else, are we missing test cases that should throw an error?

auth/auth.go
```diff
@@ -36,6 +37,7 @@ var ProviderVersion = "unknown"
 type Auth struct {
 	region, nameSpace, svcAcc, podName, preferredAddressType string
 	usePodIdentity                                           bool
+	podIdentityHttpTimeout                                   time.Duration
```
i-am-SR: nit: consider updating the struct comment above describing the new members used for initialization.

auth/auth_test.go
```diff
@@ -38,18 +39,122 @@ var sessionTests []sessionTest = []sessionTest{
...
+		},
+		{
+			name:                   "valid auth with IRSA",
+			region:                 "us-east-1",
+			nameSpace:              "kube-system",
+			svcAcc:                 "irsa-sa",
+			podName:                "irsa-pod",
+			preferredAddressType:   "ipv6",
+			usePodIdentity:         false,
+			podIdentityHttpTimeout: 200 * time.Millisecond,
```
i-am-SR: Is it valid for customers to not use PodIdentity (use IRSA) but still configure its HTTPTimeout?

auth/auth_test.go
```diff
@@ -38,18 +39,122 @@ var sessionTests []sessionTest = []sessionTest{
...
+					t.Errorf("Expected usePodIdentity %v, got %v", tt.usePodIdentity, auth.usePodIdentity)
+				}
+				if auth.podIdentityHttpTimeout != tt.podIdentityHttpTimeout {
+					t.Errorf("Expected podIdentityHttpTimeout %v, got %v", tt.podIdentityHttpTimeout, auth.podIdentityHttpTimeout)
+				}
+				if auth.k8sClient == nil {
+					t.Error("Expected k8sClient to be set")
+				}
+				if auth.stsClient == nil {
+					t.Error("Expected stsClient to be set")
```
i-am-SR, line 138: As per the logic in auth, stsClient is only initialized when using PodIdentity. In the test case 2 (using IRSA) the stsClient is expected to be nil, right? What am I missing?

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -358,6 +359,61 @@ func TestPodIdentityToken(t *testing.T) {
 	}
 }
 
+func TestNewPodIdentityCredentialProviderTimeout(t *testing.T) {
+	tests := []struct {
+		name                   string
+		podIdentityHttpTimeout time.Duration
+		expectError            bool
```
i-am-SR, line 372: same as above: Are we missing tests for cases that will throw errors?

main.go
```diff
@@ -20,12 +21,37 @@ import (
...
 
+// parsePodIdentityHttpTimeout parses and validates the HTTP timeout for Pod Identity authentication
+func parsePodIdentityHttpTimeout(timeoutStr string) time.Duration {
+	if timeoutStr == "" {
+		return 100 * time.Millisecond
+	}
+
+	duration, err := time.ParseDuration(timeoutStr)
+	if err != nil {
+		klog.Errorf("failed to parse podIdentityHttpTimeout value '%s': %v", timeoutStr, err)
```
i-am-SR: nit: worth mentioning in the log message - using default value 100 ms?

main.go
```diff
@@ -20,12 +21,37 @@ import (
...
+		return 100 * time.Millisecond
+	}
+
+	if duration <= 0 {
+		klog.Errorf("podIdentityHttpTimeout must be positive, got: %v", duration)
```
i-am-SR: same.

main_test.go
```diff
@@ -0,0 +1,299 @@
...
+			var hasPositiveError bool
+			var hasHighValueWarning bool
+
+			if tt.timeoutValue != "" {
+				podIdentityHttpTimeoutDuration, parseErr = time.ParseDuration(tt.timeoutValue)
+				if parseErr != nil && tt.expectError {
+					// Expected parse error
+				} else if parseErr != nil {
+					t.Errorf("Unexpected parse error: %v", parseErr)
+				}
```
i-am-SR:
```
if parseErr != nil && !tt.expectError {
		t.Errorf("Unexpected parse error: %v", parseErr)
}
```
to simplify?

main_test.go
```diff
@@ -0,0 +1,299 @@
...
+			var podIdentityHttpTimeoutDuration time.Duration
+			var parseErr error
+			var hasPositiveError bool
+			var hasHighValueWarning bool
+
+			if tt.timeoutValue != "" {
+				podIdentityHttpTimeoutDuration, parseErr = time.ParseDuration(tt.timeoutValue)
+				if parseErr != nil && tt.expectError {
+					// Expected parse error
+				} else if parseErr != nil {
```
i-am-SR: We also default to 100 ms if there was a parse error, right?

main_test.go
```diff
@@ -0,0 +1,299 @@
...
+				// Default to 100ms
+				podIdentityHttpTimeoutDuration = 100 * time.Millisecond
+			}
+
+			// Verify the expected result
+			if !tt.expectError || parseErr == nil {
+				if podIdentityHttpTimeoutDuration != tt.expectedResult {
+					t.Errorf("Expected timeout %v, got %v", tt.expectedResult, podIdentityHttpTimeoutDuration)
+				}
+			}
```
i-am-SR: if we update the `podIdentityHttpTimeoutDuration` to the default value in line #156, this can be simplified to:
```
if podIdentityHttpTimeoutDuration != tt.expectedResult {
	t.Errorf("Expected timeout %v, got %v", tt.expectedResult, podIdentityHttpTimeoutDuration)
}
```
right?

main_test.go
```diff
@@ -0,0 +1,299 @@
...
+			if tt.expectError && parseErr == nil && !hasPositiveError {
+				t.Errorf("Expected an error but got none")
+			}
+
+			// Verify warning conditions
+			if tt.expectWarning != hasHighValueWarning {
+				t.Errorf("Expected warning: %v, got warning: %v", tt.expectWarning, hasHighValueWarning)
+			}
+		})
+	}
```
i-am-SR: To me this logic of using booleans is not extensible and it complicates things. IMU the only thing we care to validate is the value in `podIdentityHttpTimeoutDuration` matches the value in `tt.expectedResult`. 

Also, I feel the test function `TestParsePodIdentityHttpTimeout` already validates the parsing logic. Do we really need these tests and if so, can we simplify this test?

main_test.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func TestPodIdentityHttpTimeoutValidation(t *testing.T) {
```
i-am-SR: same comment: what is the value of testing this?

auth/auth_test.go
ThirdEyeSqueegee: No but we need to specify it to construct an `Auth` struct, I'll change it to `100 * time.Millisecond` for clarity

auth/auth_test.go
ThirdEyeSqueegee: It is but `expectError` is also false there. This is just boilerplate for later when we add negative test cases.

charts/secrets-store-csi-driver-provider-aws/values.yaml
```diff
@@ -45,6 +45,8 @@ securityContext:
 
 useFipsEndpoint: false
 
+podIdentityHttpTimeout: 100ms
```
junpengdev: Add a note here to increase the timeout based on our discussion.

simonmarty:
@junpengdev We're updating the default to match the default behavior of the AWS SDK for Go v2. The default timeout in their client is 30s which should be plenty for most customers.

https://github.com/aws/aws-sdk-go-v2/blob/main/aws/transport/http/client.go#L19-L34

crus-umich (approved):

PR #475, opened 2025-06-18
fix: makefile tabs and `docker-login` target

ThirdEyeSqueegee:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,8 @@ $(eval BUILD_DATE=$(shell date -u +%Y.%m.%d.%H.%M))
 FULL_REV=$(MAJOR_REV).$(MINOR_REV).$(PATCH_REV)-$(BUILD_DATE)
 
 LDFLAGS?="-X github.com/aws/secrets-store-csi-driver-provider-aws/server.Version=$(FULL_REV) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,38 +32,38 @@ CHART_RELEASER_PATH ?= cr
...
 
 docker-login:
-    # Logging into ecr-public is required to pull the Amazon Linux 2 image used for the build
-    aws --region us-east-1 ecr-public get-login-password | docker login -u AWS --password-stdin public.ecr.aws
-    ifneq ($(REPOBASE), public.ecr.aws)
-        aws --region $(AWS_REGION) ecr get-login-password | docker login -u AWS --password-stdin $(REPOBASE)
-    endif
+	# Logging into ecr-public is required to pull the Amazon Linux 2 image used for the build
+	aws --region us-east-1 ecr-public get-login-password | docker login -u AWS --password-stdin public.ecr.aws
+	@if [[ "$(REPOBASE)" != "public.ecr.aws" ]]; then\
```
simonmarty, line 44: Why the change from ifneq/ifeq to `@if`?

ThirdEyeSqueegee: `ifneq` doesn't work in targets, only at the top level. So we have to use regular bash conditionals

simonmarty (approved):

PR #488, opened 2025-07-10
Implement auto-fallback for IPv4/IPv6 in PodIdentityCredentialProvider

vedant-jaiswal:
*Description of changes:*
- Added parseAddressPreference helper to normalize empty input to "auto"
- Refactored GetAWSConfig method to use the new helper
- Implemented IPv4 to IPv6 fallback for "auto" preference
- Maintained explicit IPv4 and IPv6 options

### DIFF
--- a/credential_provider/pod_identity_credential_provider.go
+++ b/credential_provider/pod_identity_credential_provider.go
@@ -113,27 +113,30 @@ func parseAddressPreference(preferredAddressType string) string {
 }
 
 func (p *PodIdentityCredentialProvider) GetAWSConfig(ctx context.Context) (aws.Config, error) {
+	var configErr error
 	preference := parseAddressPreference(p.preferredAddressType)
 	if preference == "auto" || preference == "ipv4" {
 		config, err := p.getConfigWithEndpoint(ctx, podIdentityAgentEndpointIPv4)
 		if err != nil {
 			klog.Warningf("IPv4 endpoint attempt failed: %v", err)
+			configErr = err
 		} else {
 			return config, nil
 		}
 	}
 
-	if preference == "ipv6" || preference == "auto" {
+	if preference == "auto" || preference == "ipv6" {
 		config, err := p.getConfigWithEndpoint(ctx, podIdentityAgentEndpointIPv6)
 
 		if err != nil {
-			klog.Warningf("IPv4 endpoint attempt failed: %v", err)
+			klog.Warningf("IPv6 endpoint attempt failed: %v", err)
+			configErr = err
 		} else {
 			return config, nil
 		}
 	}
 
-	return aws.Config{}, fmt.Errorf("failed to get AWS config %s", p.preferredAddressType)
+	return aws.Config{}, fmt.Errorf("failed to get AWS config from pod identity agent: %+v", configErr)
 }
 
 func (p *PodIdentityCredentialProvider) getConfigWithEndpoint(ctx context.Context, endpoint string) (aws.Config, error) {
--- a/credential_provider/pod_identity_credential_provider.go
+++ b/credential_provider/pod_identity_credential_provider.go
@@ -2,10 +2,10 @@ package credential_provider
 
 import (
 	"context"
+	"fmt"
+	"k8s.io/klog/v2"
 	"net/http"
 	"time"
-	"fmt"
-    "k8s.io/klog/v2"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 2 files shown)


credential_provider/pod_identity_credential_provider.go
```diff
@@ -67,33 +71,73 @@ func (p *podIdentityTokenFetcher) GetToken() (string, error) {
...
+		config, err := p.getConfigWithEndpoint(ctx, podIdentityAgentEndpointIPv6)
+
+		if err != nil {
+			klog.Warningf("IPv4 endpoint attempt failed: %v", err)
+		} else {
+			return config, nil
+		}
+	}
+
+	return aws.Config{}, fmt.Errorf("invalid preferred address type: %s", p.preferredAddressType)
```
simonmarty: The error message here is incorrect. The preferred address type cannot be invalid given the above.

It should be something along the lines of "failed to get config"

simonmarty (changes requested):

credential_provider/pod_identity_credential_provider.go
vedant-jaiswal: Yes, changing the error message.

credential_provider/pod_identity_credential_provider.go
```diff
@@ -67,33 +71,73 @@ func (p *podIdentityTokenFetcher) GetToken() (string, error) {
...
 }
 
 func NewPodIdentityCredentialProvider(
 	region, nameSpace, svcAcc, podName, preferredAddressType string,
 	k8sClient k8sv1.CoreV1Interface,
 ) (ConfigProvider, error) {
-	endpoint := podIdentityAgentEndpointIPv4
-	if preferredAddressType == "ipv6" {
-		endpoint = podIdentityAgentEndpointIPv6
+	// Add validation if needed
```
simonmarty, line 84: Why were those lines added?

credential_provider/pod_identity_credential_provider.go
```diff
@@ -67,33 +71,73 @@ func (p *podIdentityTokenFetcher) GetToken() (string, error) {
...
+		config, err := p.getConfigWithEndpoint(ctx, podIdentityAgentEndpointIPv6)
+
+		if err != nil {
+			klog.Warningf("IPv4 endpoint attempt failed: %v", err)
+		} else {
+			return config, nil
+		}
+	}
+
+	return aws.Config{}, fmt.Errorf("failed to get AWS config %s", p.preferredAddressType)
```
simonmarty: It looks like this error differs from the error we had before, https://github.com/aws/secrets-store-csi-driver-provider-aws/commit/529f812842b350baf1ae599fd7a01cda01415254#diff-334d3897f8b712366b775ee92873760fe62385d60cedf2d025198f5bbfce0540L156

Can we keep as similar as to what it was before as possible? That way we minimize backwards incompatible changes in the next release.

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -59,20 +88,27 @@ func newPodIdentityCredentialWithMock(t *testing.T, isIPv4 bool, tstData podIden
...
+		if isIPv4 {
+			podIdentityAgentEndpointIPv4 = "http://127.0.0.1:1"
+		} else {
+			podIdentityAgentEndpointIPv6 = "http://[::1]:1"
+		}
 	} else {
-		credentialEndpoint = mockServer.URL
-		podIdentityAgentEndpointIPv6 = mockServer.URL
+		if isIPv4 {
+			podIdentityAgentEndpointIPv4 = mockServer.URL
```
simonmarty, line 100: Why this change?

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -89,38 +125,61 @@ var podIdentityCredentialTests []podIdentityCredentialTest = []podIdentityCreden
...
 		t.Run(tstData.testName, func(t *testing.T) {
-			provider, closer := newPodIdentityCredentialWithMock(t, tstData.preferredEndpoint == "ipv4", tstData)
-			defer func() { closer <- struct{}{} }()
+			isIPv4 := tstData.preferredEndpoint == "ipv4" ||
+				(tstData.preferredEndpoint == "auto" && !tstData.podIdentityError)
 
-			cfg, _ := provider.GetAWSConfig(context.Background())
-			_, err := cfg.Credentials.Retrieve(context.Background())
+			provider, closer := newPodIdentityCredentialWithMock(t, isIPv4, tstData)
+			defer func() { closer <- struct{}{} }()
```
simonmarty: Why is this needed?

simonmarty (changes requested):

credential_provider/pod_identity_credential_provider.go
vedant-jaiswal: Yes, making it same to the error we had previously.

credential_provider/pod_identity_credential_provider.go
simonmarty: Can you add tests for the validation below?

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -89,38 +125,61 @@ var podIdentityCredentialTests []podIdentityCredentialTest = []podIdentityCreden
 	{"Pod Identity Success via IPv4", false, false, false, "ipv4", ""},
 	{"Pod Identity Success via IPv6", false, false, false, "ipv6", ""},
 	{"Pod Identity Success via auto selection", false, false, false, "auto", ""},
-	{"Pod identity Failure via IPv4", false, false, true, "ipv4", "failed to refresh cached credentials, failed to load credentials, exceeded maximum number of attempts, 3, : "},
-	{"Pod identity Failure via IPv6", false, false, true, "ipv6", "failed to refresh cached credentials, failed to load credentials, exceeded maximum number of attempts, 3, : "},
-	{"Pod identity Failure via auto selection", false, false, true, "auto", "failed to refresh cached credentials, failed to load credentials, exceeded maximum number of attempts, 3, : "},
+	{"Pod identity Failure via IPv4", false, false, true, "ipv4", "failed to refresh cached credentials"},
```
simonmarty: As much as I think this is a fine approach for `TestParseAddressPreference` since it fits in about 20 lines, writing tests like this starts getting confusing when you have multiple booleans/test cases that diverge too much.

Don't be afraid of being wordy in tests. You can separate your test cases in multiple different tests where you perform
- Setup: configure your state to trigger a specific outcome (in your case, set up ipv4/ipv6 endpoint behavior depending on the test case. Ex: for `auto`, you can have one test case where the ipv4 is routed to a no-op and ipv6 is routed to fake creds to test fallback)
- Action: perform the action the test is checking (in your case, instantiate the creds provider and call it)
- Validate: check the output of the action for expected state.

credential_provider/pod_identity_credential_provider.go
vedant-jaiswal: Sure!

credential_provider/pod_identity_credential_provider_test.go
vedant-jaiswal: Yes makes sense, lets change from table-driven testing to individual cases test function.

ThirdEyeSqueegee (approved):

PR #495, opened 2025-07-23
Add note about grpcSupportedProviders flag for older driver versions

vedant-jaiswal:
*Description of changes:*

Add documentation note that older versions of the Secrets Store CSI driver may require 
the `--set grpcSupportedProviders="aws"` flag during installation. 
Also clarify that this step can be skipped when installing via Helm, as the chart automatically installs 
a compatible CSI driver version by default, which can be disabled with 
`secrets-store-csi-driver.install=false.`

PR: https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/384

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ AWS offers two services to manage secrets and parameters conveniently in your co
     helm repo add secrets-store-csi-driver https://kubernetes-sigs.github.io/secrets-store-csi-driver/charts
     helm install -n kube-system csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver
     ```
-  **Note** that older versions of the driver may require the ```--set grpcSupportedProviders="aws"``` flag on the install step.
+  **Note** that older versions of the driver may require the ```--set grpcSupportedProviders="aws"``` flag on the install step.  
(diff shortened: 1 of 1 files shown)


simonmarty (approved):

README.md
```diff
@@ -14,7 +14,8 @@ AWS offers two services to manage secrets and parameters conveniently in your co
     helm repo add secrets-store-csi-driver https://kubernetes-sigs.github.io/secrets-store-csi-driver/charts
     helm install -n kube-system csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver
     ```
-  **Note** that older versions of the driver may require the ```--set grpcSupportedProviders="aws"``` flag on the install step.
+  **Note** that older versions of the driver may require the ```--set grpcSupportedProviders="aws"``` flag on the install step.  
+  **Note** this step can be skipped if installing from Helm. The Helm chart for the ASCP by default automatically installs a compatible version of the Secrets Store CSI driver as a helm dependency. This can be disabled by setting `secrets-store-csi-driver.install=false`.
```
i-am-SR, line 18: nit: Note that*

PR #523, opened 2025-09-23
Add helm chart publishing action

simonmarty:
*Issue #, if available:*

*Description of changes:* Add a manually triggered github action to publish the helm chart.

### DIFF
--- a/.github/workflows/release-chart.yml
+++ b/.github/workflows/release-chart.yml
@@ -28,7 +28,7 @@ jobs:
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
       - name: Setup Helm
-        uses: Azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4
+        uses: Azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # v4.3.1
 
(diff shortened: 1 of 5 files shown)


ThirdEyeSqueegee (dismissed):

.github/workflows/release-chart.yml
```diff
@@ -0,0 +1,42 @@
...
+    uses: ./.github/workflows/go.yml
+  release:
+    needs: [test]
+    permissions:
+      contents: write
+    environment: prod
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
```
rstevens011, line 17: Why are these fixed versions?

simonmarty: Tags in github are not immutable by default, there's some WIP to address that https://docs.github.com/en/actions/how-tos/create-and-publish-actions/using-immutable-releases-and-tags-to-manage-your-actions-releases

The concern is a supply chain attack where a threat actor points the newest version tag to code containing a malicious payload.

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

PR #539, opened 2025-10-17
Add integration tests workflow

simonmarty:
*Issue #, if available:*

*Description of changes:* Fork run: https://github.com/simonmarty/secrets-store-csi-driver-provider-aws/actions/runs/18887954508 (will try to keep this link up to date during review)

Add initial integration test workflow. The workflow triggers on `pull_request_target` similar to https://github.com/aws/aws-secretsmanager-agent/pull/136

### DIFF
--- a/charts/secrets-store-csi-driver-provider-aws/templates/_helpers.tpl
+++ b/charts/secrets-store-csi-driver-provider-aws/templates/_helpers.tpl
@@ -41,6 +41,9 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app: {{ template "provider.name" . }}
+{{- with .Values.commonLabels }}
+{{ toYaml . }}
+{{- end }}
 {{- end }}
 
 {{/*
--- a/charts/secrets-store-csi-driver-provider-aws/values.yaml
+++ b/charts/secrets-store-csi-driver-provider-aws/values.yaml
@@ -11,6 +11,7 @@ kubeletPath: "/var/lib/kubelet"
 
 driverWritesSecrets: false
 
+commonLabels: {}
 podLabels: {}
 podAnnotations: {}
 


.github/workflows/integ.yml
```diff
@@ -0,0 +1,59 @@
...
+        include:
+          - os: linux
+          - arch: amd64
+            runner: ubuntu-latest
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+
+    steps:
+      - name: Install parallel
```
ThirdEyeSqueegee: I don't think you need this since you're using a `matrix`, the `Run integ tests` step will run each matrix entry in parallel anyway

simonmarty: Yeah, plus parallel seems to be already installed on the workers anyways.

tests/run-tests.sh
```diff
@@ -30,41 +35,57 @@ cleanup_generated_files() {
...
+		delete_cluster integ-cluster-arm-pod-identity
+	fi
+	if [[ "$2" == "all" || "$2" == "arm" || "$2" == "irsa" || "$2" == "arm-irsa" ]]; then
+		delete_cluster integ-cluster-arm-irsa
+	fi
+
+	exit $?
+fi
+
+check_parallel
```
ThirdEyeSqueegee: I don't think we want to factor out `check_parallel` here, we don't care if `parallel` is installed if a single config is provided to `run-tests` (i.e., `./run-tests x64-pod-identity`).

.github/workflows/integ.yml
```diff
@@ -0,0 +1,135 @@
...
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3"
+
+      - name: Install Python dependencies
+        run: pip install boto3
+
+      - name: Setup Bats and bats libs
+        id: setup-bats
+        uses: bats-core/bats-action@3.0.1
```
ThirdEyeSqueegee, line 116: Does `@3` not work here?

tests/run-tests.sh
```diff
@@ -30,14 +35,33 @@ cleanup_generated_files() {
...
+}
+
 cleanup_secrets() {
 	echo "Cleaning up secrets and parameters..."
 	python3 generate-test-files.py cleanup-secrets
 }
 
-# Trap to ensure cleanup happens even if script is interrupted
-trap cleanup_generated_files EXIT
-trap cleanup_secrets EXIT
```
ThirdEyeSqueegee, line 40: How come we're getting rid of these?

.github/workflows/integ.yml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: List images
+        run: |
+          docker image ls -a
+  integration-tests:
+    name: Run Integration Tests
+    needs: build-docker
+    if: |
+      github.event_name != 'pull_request_target' ||
+      github.event.pull_request.author_association == 'MEMBER' ||
```
reyhankoyun: Do we still need the MEMBER check if we're not going to trigger the workflow automatically based on authorship?

simonmarty: Yep I can remove it now thanks.

.github/workflows/integ.yml
simonmarty: They don't publish major version tags/releases https://github.com/bats-core/bats-action/tags

tests/run-tests.sh
simonmarty: They trap and overwrite the exit code of the workflow which prevents accurate reporting of workflow success/failures to the github actions runtime. I've replaced these with an `always()` block in the GH actions workflow. I haven't found a better way to do this so far.

.github/workflows/pr_sync.yml
```diff
@@ -0,0 +1,31 @@
+name: Remove safe-to-test label when new commits are pushed
+
+on:
+  pull_request:
```
simonmarty: Double checking that the OIDC token provided to this workflow has permissions to remove tags on the PR when the PR author is not a maintainer of the destination repo.

.github/workflows/integ.yml
```diff
@@ -0,0 +1,138 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches: ["main"]
+  pull_request_target:
+    types: [labeled]
+
+concurrency:
+  # Only run one workflow at a time to avoid hitting the network interface quota
+  group: integration-tests
```
reyhankoyun, line 12: Is this going to make it so that only one integration test can run across the entire repository at any time? For example, if PR #1 starts tests, should PR #2 wait? If so, should we consider per-PR isolation so that each PR gets its own concurrency group? Something like:

```
concurrency:
  group: integration-tests-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```
[Ref](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#:~:text=branches%3A%0A%20%20%20%20%20%20%2D%20main-,concurrency%3A%0A%20%20group%3A%20%24%7B%7B%20github.workflow%20%7D%7D%2D%24%7B%7B%20github.ref%20%7D%7D%0A%20%20cancel%2Din%2Dprogress%3A%20true,-You%20can%20also)

simonmarty: From what I understand from the docs linked above, the `group` field is a key to a lock. What I want is a global lock across all possible executions of the workflow. If I specify a key that includes the pull request number like you linked above, that means that two workflow runs with group ids `integration-tests-1` and `integration-tests-2` could run concurrently. The workflow would then both attempt to create EKS clusters and exceed the NAT gateway quota limit in the account.

simonmarty:
This is not necessary, the pull request gets squashed on merge with the pull request title and description as the squashed commit title and description header respectively.

.github/workflows/integ.yml
```diff
@@ -0,0 +1,138 @@
...
+  build-docker:
+    name: Build Docker Image
+    if: |
+      github.event_name != 'pull_request_target' ||
+      contains(github.event.pull_request.labels.*.name, 'safe-to-test')
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
```
lohdipak, line 25: Is it correct or we need something like this. in current code we are running ubuntu-24.04-arm  on amd as well.

matrix:
    include:
      - runner: ubuntu-latest
        arch: amd64
      - runner: ubuntu-24.04-arm  
        arch: arm64

.github/workflows/integ.yml
```diff
@@ -0,0 +1,138 @@
...
+    if: |
+      github.event_name != 'pull_request_target' ||
+      contains(github.event.pull_request.labels.*.name, 'safe-to-test')
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
```
lohdipak, line 72: Are we planning to run the integ test against 8 combination? can we optimized this.

simonmarty: 4 combinations (cardinality of `runner` * cardinality of `auth_type`). It might be possible to reuse the same EKS cluster for `irsa` and `pod-identity` for a given arch. I will investigate in a subsequent PR.

.github/workflows/integ.yml
simonmarty: Here's a run https://github.com/simonmarty/secrets-store-csi-driver-provider-aws/actions/runs/18793931196

The purpose of the include field is described here https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#expanding-or-adding-matrix-configurations

The number of configurations is actually the cardinality of `runner`.

.github/workflows/integ.yml
ThirdEyeSqueegee: Cluster reuse prolly doesn't matter since we're doing everything in parallel anyway, no?

simonmarty: It's a tad more resource efficient. These clusters are short lived though. They get torn down after test execution ends.

PR #541, opened 2025-11-07
test: add EKS add-on testing support

ThirdEyeSqueegee:
*Issue #, if available:*
N/A

*Description of changes:*
* Rewrite integration tests with a new `run-tests.sh` runner and `test-manager.py` for managing test secrets/parameters
* Add `--addon` flag to install the provider as an EKS add-on (driver bundled), with optional `--version` flag
* Add `setup.sh` for local dev environment setup (bats, Python venv, IAM role)
* Add per-target parallelism, preflight checks (tools, credentials, VPC capacity, stale resource cleanup), and fail-fast diagnostics
* Add `BasicTestMountSPC.yaml.template` and `BasicTestMount.yaml.template` driven entirely by env vars (no file generation step)
* Remove old integration test files

*Testing:*
Tests pass locally.

### DIFF
--- a/deployment/private-installer.yaml
+++ b/deployment/private-installer.yaml
@@ -56,8 +56,6 @@ spec:
     spec:
       serviceAccountName: csi-secrets-store-provider-aws
       hostNetwork: false
-      imagePullSecrets:
-        - name: ghcr-credentials
       containers:
         - name: provider-aws-installer
           image: ${PRIVREPO}${PRIVTAG:+:}${PRIVTAG}
--- a/tests/integration.bats
+++ b/tests/integration.bats
@@ -207,7 +207,14 @@ validate_jmes_mount() {
 	log "Installing AWS provider"
 
 	local image="${PRIVREPO}${PRIVTAG:+:${PRIVTAG}}"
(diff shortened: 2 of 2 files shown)


deployment/private-installer.yaml
```diff
@@ -56,6 +56,8 @@ spec:
     spec:
       serviceAccountName: csi-secrets-store-provider-aws
       hostNetwork: false
+      imagePullSecrets:
+        - name: ghcr-credentials
```
simonmarty: Is that gonna break usages of the private installer that don't use this secret?

tests/integration.bats
```diff
@@ -104,6 +104,15 @@ setup_file() {
 	install_driver
 	setup_auth
 
+	# Create ghcr.io pull secret for private test images
+	if [[ -n "${GHCR_TOKEN:-}" ]]; then
+		log "Creating ghcr.io image pull secret"
+		kctl create secret docker-registry ghcr-credentials \
```
simonmarty, line 110: What's this `kctl`? `kubectl`?

ThirdEyeSqueegee: Yes, just a wrapper for `kubectl`

deployment/private-installer.yaml
ThirdEyeSqueegee: It'll emit a warning but won't cause any failures, I can make it conditional tho let me push another commit

PR #544, opened 2025-11-15
chore: fetch nginx from ECR Public Gallery

ThirdEyeSqueegee:
*Issue #, if available:*
N/A

*Description of changes:*
- The nginx image currently defined in the example deployment files isn't accessible from clusters in AWS China regions. Fetch nginx from the ECR Public Gallery to fix this.
- Format yaml files

### DIFF
--- a/examples/ExampleDeployment-PodIdentity.yaml
+++ b/examples/ExampleDeployment-PodIdentity.yaml
@@ -38,7 +38,7 @@ spec:
               secretProviderClass: "nginx-pod-identity-deployment-aws-secrets"
       containers:
         - name: nginx-pod-identity-deployment
-          image: nginx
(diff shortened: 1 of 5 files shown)


examples/ExampleDeployment-PodIdentity.yaml
```diff
@@ -38,10 +38,10 @@ spec:
               secretProviderClass: "nginx-pod-identity-deployment-aws-secrets"
       containers:
         - name: nginx-pod-identity-deployment
-          image: nginx
+          image: public.ecr.aws/nginx/nginx:latest
```
simonmarty, line 41: There's really only two lines of diff in here, can we remove the formatting changes?

PR #545, opened 2025-11-17
Remove unneeded helpers, use kubectl waiters, set bash strict mode

simonmarty:
*Issue #, if available:*

*Description of changes:*
- `wait_for_process` isn't needed as far as I can tell. We can do all the waiting we need using `kubectl`'s `wait` command.
- Set bash strict mode to exit non-zero on subcommand failure. This ensures we exit non-zero on a failing bats test.

### DIFF
--- a/.github/workflows/release-chart.yml
+++ b/.github/workflows/release-chart.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
 
 permissions:
+  contents: read
   id-token: write
 
 jobs:


tests/integration.bats.template
```diff
@@ -2,8 +2,6 @@
 
 load helpers
 
-WAIT_TIME=120
-SLEEP_TIME=1
```
reyhankoyun, line 6: Could we add these variables back in? I think these are needed by `wait_for_process` helper function.

simonmarty: Ah, this made me realize I can't strip out `wait_for_process` because it's being used by the `aws.bats` file used by the upstream driver end to end tests, putting those back in.

The two env vars can be stripped out though, the usage of these helpers has been removed.

ThirdEyeSqueegee: The driver doesn't actually use our `aws.bats` file, theirs diverged a while ago. I removed `aws.bats` and the rest of the old testing setup in #541

PR #547, opened 2025-11-20
ci: add `paths-ignore` to `integ.yml`

ThirdEyeSqueegee:
*Issue #, if available:*
N/A

*Description of changes:*
Exclude paths for `integ.yml`

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches: ["main"]
     paths-ignore:
-      - ".github/**"
       - ".*ignore"
       - "*.md"
       - "LICENSE"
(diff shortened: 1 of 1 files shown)


.github/workflows/integ.yml
```diff
@@ -4,8 +4,18 @@ on:
   workflow_dispatch:
   push:
     branches: ["main"]
+    paths-ignore:
+      - ".*ignore"
```
simonmarty, line 8: This is going to be a problem because merging a PR requires the integ workflow to pass.

ThirdEyeSqueegee: Ah I see, thoughts about relaxing that for the given file types? I think we can safely avoid running integ tests for changes in such files

simonmarty: I don't believe there's a way to have GitHub conditionally add PR required checks.

PR #555, opened 2026-01-07
Support for AssumeRole in AWS Secrets Store CSI Provider for `SecretProviderClass`

dvp34:
## Summary

This PR adds support for assuming an IAM role at mount time via a new `SecretProviderClass` parameter: `assumeRoleArn`. When specified, the AWS provider will perform an STS `AssumeRole` using the base credentials (IRSA or Pod Identity) and use the assumed-role credentials for all AWS Secrets Manager and SSM calls for that mount.

Optional parameters are also supported:
- `assumeRoleDurationSeconds`
- `assumeRoleExternalId`

This enables cleaner and more scalable cross-account access patterns for secrets.

---

## Problem Statement

Currently, `secrets-store-csi-driver-provider-aws` retrieves secrets using the credentials obtained via IRSA or Pod Identity and directly calls AWS Secrets Manager or SSM with those credentials.

In cross-account setups (for example, EKS in Account A accessing Secrets Manager in Account B), this requires:

- The IRSA role to have `secretsmanager:GetSecretValue` permissions, and
- A resource policy on each secret in Account B explicitly allowing access from the IRSA role.

This approach introduces significant operational overhead, especially when multiple workloads or clusters need access to cross-account secrets.

Although users may specify an `assumeRoleArn` (or similar) under `spec.parameters` today, the AWS provider does not parse or act on this value. No `sts:AssumeRole` call is made, and the parameter is effectively ignored. As a result, common cross-account access patterns are not supported.

---

## Proposed Solution

Add support for assuming a role per mount by introducing the following `SecretProviderClass` parameters under `spec.parameters`:

- `assumeRoleArn` (string, required)  
  Instructs the provider to call STS `AssumeRole` using the base credentials (IRSA or Pod Identity).

- `assumeRoleDurationSeconds` (string or int, optional)  
  Session duration for the assumed role.

- `assumeRoleExternalId` (string, optional)  
  External ID to pass to the `AssumeRole` call.

When `assumeRoleArn` is specified, the IRSA role only needs permission to call `sts:AssumeRole`. All Secrets Manager or SSM access is performed using the assumed role, eliminating the need for:

- Direct Secrets Manager permissions on the IRSA role, and
- Resource policies on secrets for each consuming role.

---

## Example `SecretProviderClass`

```yaml
apiVersion: secrets-store.csi.x-k8s.io/v1
kind: SecretProviderClass
metadata:
  name: cross-account-secrets
spec:
  provider: aws
  parameters:
    region: us-east-1
    assumeRoleArn: arn:aws:iam::123456789012:role/target-role
    assumeRoleDurationSeconds: "900"
    objects: |
      - objectName: "arn:aws:secretsmanager:us-east-1:123456789012:secret:MySecret-abc123"
        objectType: "secretsmanager"
```
---

## Behaviour & implementation notes
- When `assumeRoleArn` is present:
  1. Provider obtains base credentials via IRSA or Pod Identity (current behavior).
  2. Provider calls STS AssumeRole (with optional Duration/ExternalId) using those credentials.
  3. Provider uses the resulting assumed-role creds to instantiate region clients (Secrets Manager / SSM) for the mount request.
- If `assumeRoleArn` is missing, current behaviour is unchanged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,9 @@ RUN echo "Running on ${BUILDPLATFORM}, building for ${TARGETPLATFORM}."
 
 WORKDIR /workdir
 
-RUN apk add git build-base
-RUN go env -w GOPROXY=direct
+RUN apk add --no-cache git build-base ca-certificates
+RUN update-ca-certificates || true
+RUN go env -w GOPROXY=https://proxy.golang.org,direct
 
 COPY go.mod .
 COPY go.sum .
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,8 @@ RUN echo "Running on ${BUILDPLATFORM}, building for ${TARGETPLATFORM}."
 
 WORKDIR /workdir
 
-RUN apk add --no-cache git build-base ca-certificates
-RUN update-ca-certificates || true
-RUN go env -w GOPROXY=https://proxy.golang.org,direct
+RUN apk add --no-cache git build-base
+RUN go env -w GOPROXY=direct
 
 COPY go.mod .
 COPY go.sum .
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -72,19 +72,19 @@ func NewAuth(
 	}
 
 	return &Auth{
-		region:                 region,
-		nameSpace:              nameSpace,
-		svcAcc:                 svcAcc,
-		podName:                podName,
-		preferredAddressType:   preferredAddressType,
-		eksAddonVersion:        eksAddonVersion,
-		usePodIdentity:         usePodIdentity,
-		podIdentityHttpTimeout: podIdentityHttpTimeout,
-		k8sClient:              k8sClient,
-		stsClient:              stsClient,
-		assumeRoleArn:          assumeRoleArn,
+		region:                    region,
+		nameSpace:                 nameSpace,
+		svcAcc:                    svcAcc,
+		podName:                   podName,
+		preferredAddressType:      preferredAddressType,
+		eksAddonVersion:           eksAddonVersion,
+		usePodIdentity:            usePodIdentity,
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -5,8 +5,9 @@ RUN echo "Running on ${BUILDPLATFORM}, building for ${TARGETPLATFORM}."
 
 WORKDIR /workdir
 
-RUN apk add git build-base
-RUN go env -w GOPROXY=direct
+RUN apk add --no-cache git build-base ca-certificates
+RUN update-ca-certificates || true
```
simonmarty: What's the need behind the changes here?

auth/auth.go
```diff
@@ -41,6 +42,9 @@ type Auth struct {
 	podIdentityHttpTimeout                                                    *time.Duration
 	k8sClient                                                                 k8sv1.CoreV1Interface
 	stsClient                                                                 stscreds.AssumeRoleWithWebIdentityAPIClient
+	assumeRoleArn                                                             string
+	assumeRoleDurationSeconds                                                 string
```
simonmarty: Should this be a `time.Duration`?

Dockerfile
dvp34: I had problems when building locally, and had to have this. to fix this error
```
178.6 go: go.yaml.in/yaml/v2@v2.4.2: unrecognized import path "go.yaml.in/yaml/v2": https fetch: Get "https://go.yaml.in/yaml/v2?go-get=1": tls: failed to verify certificate: x509: certificate signed by unknown authority
------
Dockerfile:14
--------------------
  12 |     COPY go.sum .
  13 |
  14 | >>> RUN go mod download -x
  15 |
  16 |     COPY . .
--------------------
```
 I can remove this in the change.

PR #556, opened 2026-01-08
fix: use SDK AppID for User-Agent

ThirdEyeSqueegee:
*Issue #, if available:*
N/A

*Description of changes:*
The old User-Agent middleware logic was bugged causing the provider and add-on versions to not show up in service logs. This was due to our custom middleware adding an entirely new header instead of modifying the existing UserAgent header which was never ingested on the service side.
- Use the SDK's `middleware` import to fix this 

Example of UserAgent after this change (from service log):
```
UserAgent=aws-sdk-go-v2/1.39.0 ua/2.1 os/linux lang/go#1.24.12 md/GOOS#linux md/GOARCH#amd64 api/secretsmanager#1.39.4 app/secrets-store-csi-driver-provider-aws-2.1.0 m/E,z
```

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -178,7 +178,7 @@ func TestGetAWSConfig(t *testing.T) {
 	}
 }
 
-func TestGetAWSConfig_UserAgent(t *testing.T) {
+func TestAppID(t *testing.T) {
 	tests := []struct {
 		name            string
 		eksAddonVersion string
--- a/credential_provider/irsa_credential_provider_test.go
+++ b/credential_provider/irsa_credential_provider_test.go
@@ -48,6 +48,7 @@ func newIRSACredentialProviderWithMock(tstData irsaCredentialTest) *IRSACredenti
 		region:    testRegion,
 		nameSpace: testNamespace,
 		svcAcc:    testServiceAccount,
+		appID:     "test-app-id",
 		fetcher: newIRSATokenFetcher(
 			testNamespace,
 			testServiceAccount,
@@ -135,4 +136,27 @@ func TestIRSAToken(t *testing.T) {
 var irsaTokenTests []irsaCredentialTest = []irsaCredentialTest{
 	{"IRSA Token Success", false, false, "myRoleARN", true, "", ""},
 	{"IRSA Fetch JWT fail", false, true, "myRoleARN", true, "", "Fake create token"},
+}
+
+func TestNewIRSACredentialProvider_AppID(t *testing.T) {
+	expectedAppID := "test-app-id"
+	k8sClient := fake.NewSimpleClientset().CoreV1()
+
+	provider := NewIRSACredentialProvider(
+		&mockSTS{},
+		testRegion,
+		testNamespace,
+		testServiceAccount,
+		expectedAppID,
+		k8sClient,
+	)
+
+	irsaProvider, ok := provider.(*IRSACredentialProvider)
+	if !ok {
+		t.Fatal("Expected IRSACredentialProvider type")
+	}
+
+	if irsaProvider.appID != expectedAppID {
+		t.Errorf("Expected appID %q, got %q", expectedAppID, irsaProvider.appID)
+	}
 }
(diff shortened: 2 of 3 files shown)


auth/auth_test.go
```diff
@@ -180,78 +180,74 @@ func TestGetAWSConfig(t *testing.T) {
...
-				t.Errorf("HandleBuild() error = %v", err)
+				t.Fatalf("GetAWSConfig() error = %v", err)
 			}
 
-			if !nextCalled {
-				t.Error("Expected next handler to be called")
+			// Verify APIOptions are configured
+			// With eksAddonVersion: expect 2 options (provider + addon version)
+			// Without eksAddonVersion: expect 1 option (provider only)
+			expectedOptions := 1
```
i-am-SR: We test for the number of options. Can we validate the UserAgent KeyValue(s) in the Options?

auth/auth.go
```diff
@@ -105,35 +104,11 @@ func (p Auth) GetAWSConfig(ctx context.Context) (aws.Config, error) {
...
-		return stack.Build.Add(&userAgentMiddleware{
-			providerName:    ProviderName,
-			eksAddonVersion: p.eksAddonVersion,
-		}, middleware.After)
-	})
+	// Add the user agent to the config using SDK's built-in mechanism
+	cfg.APIOptions = append(cfg.APIOptions, awsmiddleware.AddUserAgentKeyValue(ProviderName, ProviderVersion))
+	if p.eksAddonVersion != "" {
+		cfg.APIOptions = append(cfg.APIOptions, awsmiddleware.AddUserAgentKeyValue("eksAddonVersion", p.eksAddonVersion))
+	}
```
ThirdEyeSqueegee: Self-review: make these mutually exclusive

auth/auth_test.go
```diff
@@ -180,78 +178,43 @@ func TestGetAWSConfig(t *testing.T) {
...
-	expectedID := "AppendCSIDriverVersionToUserAgent"
-	actualID := middleware.ID()
-
-	if actualID != expectedID {
-		t.Errorf("Expected ID() to return '%s', but got '%s'", expectedID, actualID)
-	}
-}
-
-func TestUserAgentMiddleware_HandleBuild(t *testing.T) {
+func TestGetAWSConfig_UserAgent(t *testing.T) {
```
i-am-SR: nit: this test no longer tests the GetAWSConfig method. consider renaming to something like TestAppID instead?

credential_provider/irsa_credential_provider.go
```diff
@@ -56,15 +55,15 @@ func (p *irsaTokenFetcher) GetIdentityToken() ([]byte, error) {
 
 // IRSACredentialProvider implements CredentialProvider using IAM Roles for Service Accounts
 type IRSACredentialProvider struct {
-	stsClient                 stscreds.AssumeRoleWithWebIdentityAPIClient
-	k8sClient                 k8sv1.CoreV1Interface
-	region, nameSpace, svcAcc string
-	fetcher                   stscreds.IdentityTokenRetriever
+	stsClient                        stscreds.AssumeRoleWithWebIdentityAPIClient
+	k8sClient                        k8sv1.CoreV1Interface
+	region, nameSpace, svcAcc, appID string
```
i-am-SR, line 60: Can we update the test file for this provider to use this new param and add a unit test to validate that AppID gets set in the provider? Can we add a similar test in Pod Identity CP test file too?

auth/auth_test.go
```diff
@@ -180,78 +178,43 @@ func TestGetAWSConfig(t *testing.T) {
...
 			eksAddonVersion: "v1.0.0-eksbuild.1",
-			expectedUA:      "test-provider/unknown eksAddonVersion/v1.0.0-eksbuild.1",
+			expectedAppID:   ProviderName + "-v1.0.0-eksbuild.1",
 		},
 		{
 			name:            "without EKS addon version",
-			providerName:    "test-provider",
 			eksAddonVersion: "",
-			expectedUA:      "test-provider/unknown",
+			expectedAppID:   ProviderName + "-" + ProviderVersion,
```
simonmarty, line 195: Just double checking: will this be sufficient to distinguish EKS add-on installs from regular Helm installs for the foreseeable future?

Issue #400, opened 2024-10-23
Use CSI TokenRequest for obtaining service account tokens


dronenb:
Other providers for the Secrets Store CSI driver leverage the CSI driver's (not the provider's) ability to obtain service account tokens for pods that use that CSI driver (see [here](https://kubernetes-csi.github.io/docs/token-requests.html)). This way, the providers nor the driver need the `create` `serviceaccounts/token` role.

Code in question:
https://github.com/aws/secrets-store-csi-driver-provider-aws/blob/b8df4953967406978a6199ab5321ee2308a387a7/auth/auth.go#L49-L62

Examples of this from other providers:

https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/blob/main/auth/auth.go#L187-L198

https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/41b07df322b82a56915ba63abb56c16c37730d33/pkg/auth/auth.go#L304-L327

Without this, the AWS provider will continue to be over-privileged.

jirkafajfr:
Thank you @dronenb, we've noted this as a feature request.

PR #561, opened 2026-01-30
feat: CSI Token Requests

ThirdEyeSqueegee:
*Issue #, if available:*
#400, #422 

*Description of changes:*
## Summary

This PR adopts the CSI driver's `tokenRequests` mechanism for obtaining pod service account tokens. CSI Token Requests is the standard way for CSI drivers to impersonate the pods they mount volumes for — the driver mints service account tokens on behalf of each pod and passes them to the provider. Previously, the provider minted these tokens itself using the K8s TokenRequest API, which required `serviceaccounts/token create` RBAC permission. With this change, the provider receives pre-minted tokens from the CSI driver, eliminating that permission and simplifying the provider's role to consuming tokens rather than creating them. Instead of minting Kubernetes service account tokens itself (via the TokenRequest API), the provider now receives pre-minted tokens from the CSI driver through the standard `tokenRequests` mechanism. 

## Motivation

Previously, the provider held a Kubernetes client and called the TokenRequest API directly to create service account tokens for both IRSA (`sts.amazonaws.com` audience) and Pod Identity (`pods.eks.amazonaws.com` audience). This required:

- `serviceaccounts/token` `create` permission in the ClusterRole
- A K8s client threaded through the auth layer into credential providers
- Complex mock infrastructure in tests (mock STS, mock K8s client, mock token creation)

The CSI driver already supports a `tokenRequests` feature that projects service account tokens into the mount request attributes. By leveraging this, the provider becomes a simpler consumer of pre-fetched tokens rather than an active participant in token creation.

## How it works

The CSI driver's `CSIDriver` resource is configured with `tokenRequests` specifying the required audiences:

```yaml
spec:
  tokenRequests:
    - audience: "sts.amazonaws.com"
    - audience: "pods.eks.amazonaws.com"
```

On each mount (and remount during rotation), the CSI driver mints fresh tokens via the K8s TokenRequest API and passes them to the provider in the `csi.storage.k8s.io/serviceAccount.tokens` volume attribute. The provider parses this JSON, extracts the token for the appropriate audience (IRSA or Pod Identity), and passes it to the credential provider.

**Before:**

```
Mount() → auth.NewAuth() → auth.GetAWSConfig() → credential_provider → K8s TokenRequest API → STS/PodIdentity
```

**After:**

```
Mount() → parse CSI tokens → credential_provider(token) → STS/PodIdentity
```

### Deleted files

- **`auth/auth.go`** (113 lines) — The entire auth orchestration package is removed. This contained the `Auth` struct, `NewAuth()` constructor, `GetAWSConfig()` dispatcher, `getAppID()` helper, and the `ProviderVersion`/`ProviderName` constants. All of this functionality is either inlined into the server or eliminated.
- **`auth/auth_test.go`** (221 lines) — All tests for the auth package (`TestNewAuth`, `TestGetAWSConfig`, `TestAppID`).

### New files

- **`utils/token_parser.go`** (42 lines) — Defines `ParseServiceAccountTokens()` which parses the CSI driver's `serviceAccount.tokens` JSON into a `map[string]ServiceAccountToken`, and `GetTokenForAudience()` which extracts a non-empty token for a specific audience. Also defines the `IRSAAudience` and `PodIdentityAudience` constants. Error messages explicitly point to `tokenRequests` configuration as the fix, since a missing/empty tokens attribute is the most likely misconfiguration.
- **`utils/token_parser_test.go`** (134 lines) — Table-driven tests for both functions covering: empty input, invalid JSON, valid multi-audience, valid single-audience, missing audience, empty token value, and audience constant validation.

### Application code changes

- **`server/server.go`** — The core integration point. `Mount()` now extracts `serviceAccount.tokens` from the mount attributes, calls `ParseServiceAccountTokens()` upfront, then uses `GetTokenForAudience()` to select the right token based on `usePodIdentity`. The `getRoleARN()` method (previously in `IRSACredentialProvider`) is moved here since the server already has the K8s client. The `appID()` method and `ProviderName`/`ProviderVersion` constants are moved from the deleted `auth` package. `getAwsConfigs()` signature changes to accept `roleArn` and `token` directly instead of namespace/serviceAccount/podName. A pre-existing bug in the `len(awsConfigs) > 2` guard was fixed — it previously returned `nil, err` where `err` was nil (from the preceding successful `getAwsConfigs` call), now returns a proper error.

- **`credential_provider/irsa_credential_provider.go`** — Significantly simplified. The `irsaTokenFetcher` (which called the K8s TokenRequest API) is replaced with `csiTokenFetcher`, a trivial struct that returns a pre-fetched token string to satisfy the `stscreds.IdentityTokenRetriever` interface. `NewIRSACredentialProvider()` now takes `(region, roleArn, appID, token string)` instead of `(stsClient, region, namespace, serviceAccount, appID, k8sClient)`. `getRoleARN()` is moved to the server. `GetAWSConfig()` creates an STS client inline rather than receiving one.

- **`credential_provider/pod_identity_credential_provider.go`** — Same pattern. The `podIdentityTokenFetcher` (K8s TokenRequest API with BoundObjectRef) is replaced with `csiTokenProvider`, a trivial struct returning a pre-fetched token for the `endpointcreds.AuthTokenProvider` interface. `NewPodIdentityCredentialProvider()` drops the `namespace`, `serviceAccount`, `podName`, and `k8sClient` parameters, taking `token string` instead. No more input validation for nil K8s client.

- **`credential_provider/credential_provider.go`** — Unchanged. The `ConfigProvider` interface remains the same.

- **`main.go`** — Removes the `auth` package import. Changes `auth.ProviderName` to `server.ProviderName`. Adds a startup log message: `"This provider requires tokenRequests to be configured in the CSIDriver spec"`.

### Test changes

- **`server/server_test.go`** — `buildMountReq()` now injects CSI tokens JSON into the mount attributes (both IRSA and Pod Identity tokens with far-future expiration). Error message assertions updated from `"An IAM role must be associated"` to `"IAM role must be associated"`. `auth.ProviderName` references changed to `ProviderName`. Four new tests added:
  - `TestMountMissingTokensAttribute` — verifies the error when `serviceAccount.tokens` is absent from mount attributes
  - `TestMountTokenAudienceMismatch` — verifies the error when `usePodIdentity=true` but only the IRSA token audience is present
  - `TestMountMaxRegionsExceeded` — validates the max-regions guard doesn't false-positive (and that the bug fix returns a real error)
  - `TestAppID` — table-driven test for `appID()` with and without EKS addon version override (coverage moved from deleted `auth_test.go`)

- **`credential_provider/credential_provider_test.go`** — All mock types removed (`mockSTS`, `mockK8sV1`, `mockK8sV1SA`). Only shared test constants remain.

- **`credential_provider/irsa_credential_provider_test.go`** — Rewritten from 172 to ~51 lines. Complex mock-based tests replaced with simple tests: `TestNewIRSACredentialProvider`, `TestCSITokenFetcher`, `TestIRSACredentialProvider_GetAWSConfig`.

- **`credential_provider/pod_identity_credential_provider_test.go`** — Reduced from 489 to ~100 lines. Complex mock infrastructure removed. New simple tests plus two new dual-stack tests:
  - `TestPodIdentityCredentialProvider_GetAWSConfig_AutoFallback` — verifies IPv6 works when IPv4 is unreachable
  - `TestPodIdentityCredentialProvider_GetAWSConfig_BothFail` — verifies failure when both endpoints are unreachable

### RBAC changes

- **`charts/secrets-store-csi-driver-provider-aws/templates/rbac.yaml`** — Removes the `serviceaccounts/token` `create` rule from the ClusterRole.
- **`deployment/aws-provider-installer.yaml`** — Same removal, plus adds `tolerations: [{operator: Exists}]` for EKS Auto Mode compatibility.
- **`deployment/private-installer.yaml`** — Same removal and toleration addition.

### Helm chart changes

- **`charts/secrets-store-csi-driver-provider-aws/values.yaml`** — Adds `tokenRequests` configuration to the CSI driver sub-chart dependency, specifying both audiences (`sts.amazonaws.com` and `pods.eks.amazonaws.com`).

### CI changes

- **`.github/workflows/docker-image.yml`** — Build arg changed from `auth.ProviderVersion` to `server.ProviderVersion`.
- **`.github/workflows/integ.yml`** — Same change.
- **`Makefile`** — LDFLAGS changed from `auth.ProviderVersion` to `server.ProviderVersion`.

### Documentation changes

- **`README.md`** — Adds a new "Separate CSI Driver Installation" section explaining how to configure `tokenRequests` when installing the CSI driver separately (both Helm and kubectl examples).

### Integration test changes

- **`tests/integration.bats.template`** — Adds `--set tokenRequests[0].audience=sts.amazonaws.com --set tokenRequests[1].audience=pods.eks.amazonaws.com` to the Helm install command. Adds a new test case `"Verify serviceaccounts/token create permission is not granted to provider"` that checks the ClusterRole.

## Testing

- All existing unit tests pass (adapted for the new token flow)
- 6 new unit tests added covering token parsing, error paths, and edge cases
- Integration tests pass across all 4 configurations (`arm-irsa`, `arm-pod-identity`, `x64-irsa`, `x64-pod-identity`) including rotation tests, confirming the CSI driver correctly passes fresh tokens on remount

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,7 @@ updates:
       dependencies:
         applies-to: version-updates
         update-types:
+          - major
           - minor
           - patch
   - package-ecosystem: "helm"
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -21,10 +21,10 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Build the Docker image
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           build-args: LDFLAGS=-X 'github.com/aws/secrets-store-csi-driver-provider-aws/server.Version=fakeversion' -X 'github.com/aws/secrets-store-csi-driver-provider-aws/server.ProviderVersion=fakeversion' -extldflags '-static'
           context: .
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Go
         uses: actions/setup-go@v6
@@ -57,13 +57,13 @@ jobs:
         run: go test -v ./... -coverprofile=coverage.out -covermode=atomic
 
       - name: Setup Helm
-        uses: Azure/setup-helm@v4
+        uses: Azure/setup-helm@v5
 
       - name: Helm lint
         run: helm lint charts/*
 
       - name: Codecov
-        uses: codecov/codecov-action@v5
+        uses: codecov/codecov-action@v6
         with:
           fail_ci_if_error: true
           use_oidc: true
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -33,19 +33,19 @@ jobs:
     runs-on: ${{ matrix.runner }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build the Docker image
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           build-args: LDFLAGS=-X 'github.com/aws/secrets-store-csi-driver-provider-aws/server.Version=fakeversion' -X 'github.com/aws/secrets-store-csi-driver-provider-aws/server.ProviderVersion=fakeversion' -extldflags '-static'
           context: .
@@ -83,10 +83,10 @@ jobs:
 
     steps:
       - name: Setup kubectl
-        uses: azure/setup-kubectl@v4
+        uses: azure/setup-kubectl@v5
 
       - name: Setup Helm
-        uses: azure/setup-helm@v4
+        uses: azure/setup-helm@v5
 
       - name: Setup eksctl
         run: |
@@ -113,15 +113,18 @@ jobs:
 
       - name: Setup Bats and bats libs
         id: setup-bats
-        uses: bats-core/bats-action@3.0.1
+        uses: bats-core/bats-action@4.0.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
 
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
 
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v5
+        uses: aws-actions/configure-aws-credentials@v6
         with:
           role-to-assume: ${{ secrets.ROLE_ARN }}
           role-session-name: csi-driver-ci-${{ github.run_id }}-${{ matrix.arch }}
--- a/.github/workflows/release-chart.yml
+++ b/.github/workflows/release-chart.yml
@@ -18,12 +18,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
         with:
-          egress-policy: audit
+          egress-policy: block
+          allowed-endpoints: >
+            github.com:443
+            api.github.com:443
+            uploads.github.com:443
+            kubernetes-sigs.github.io:443
+            release-assets.githubusercontent.com:443
+            get.helm.sh:443
 
       - name: Checkout
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
 
@@ -33,8 +40,7 @@ jobs:
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
       - name: Setup Helm
-        uses: Azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # v4.3.1
-
+        uses: Azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
       - name: Add SSCD Helm Repo
         run: helm repo add secrets-store-csi-driver https://kubernetes-sigs.github.io/secrets-store-csi-driver/charts
 


auth/auth.go
```diff
@@ -99,14 +89,18 @@ func (p Auth) GetAWSConfig(ctx context.Context) (aws.Config, error) {
...
+		)
 	} else {
 		klog.Infof("Using IAM Roles for Service Accounts for authentication in namespace: %s, service account: %s", p.nameSpace, p.svcAcc)
-		credProvider = credential_provider.NewIRSACredentialProvider(p.stsClient, p.region, p.nameSpace, p.svcAcc, appID, p.k8sClient)
+		credProvider, err = credential_provider.NewIRSACredentialProvider(
+			p.stsClient, p.region, p.roleArn, appID, p.serviceAccountTokens,
+		)
+	}
+
+	if err != nil {
```
simonmarty: Should we log here?

auth/auth_test.go
```diff
@@ -2,177 +2,155 @@ package auth
...
+			region:         "us-east-1",
+			usePodIdentity: true,
+			expectError:    false,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			k8sClient := fake.NewClientset().CoreV1()
+			timeout := 100 * time.Millisecond
```
simonmarty: Why the change here?

credential_provider/pod_identity_credential_provider.go
```diff
@@ -78,30 +47,34 @@ type PodIdentityCredentialProvider struct {
...
-		return nil, fmt.Errorf("k8s client cannot be nil")
+
+	token, err := utils.GetTokenForAudience(serviceAccountTokens, podIdentityAudience)
+	if err != nil {
+		klog.Errorf("Pod Identity authentication failed: could not get token for audience %q: %v", podIdentityAudience, err)
+		return nil, fmt.Errorf("failed to get Pod Identity token: %w", err)
 	}
 
-	pod_identity := PodIdentityCredentialProvider{
+	klog.V(2).Infof("Pod Identity token obtained for audience %q", podIdentityAudience)
```
simonmarty: Why is the `.V(2)` needed here?

credential_provider/pod_identity_credential_provider.go
```diff
@@ -113,31 +86,31 @@ func parseAddressPreference(preferredAddressType string) string {
 	case "ipv6":
 		return "ipv6"
 	default:
-		return "auto" // Default to auto for invalid preferences
```
simonmarty, line 116: Why was the comment removed here? Was it inaccurate?

credential_provider/pod_identity_credential_provider.go
```diff
@@ -113,31 +86,31 @@ func parseAddressPreference(preferredAddressType string) string {
...
 	}
 }
 
 func (p *PodIdentityCredentialProvider) GetAWSConfig(ctx context.Context) (aws.Config, error) {
 	var configErr error
 	preference := parseAddressPreference(p.preferredAddressType)
+
 	if preference == "auto" || preference == "ipv4" {
-		config, err := p.getConfigWithEndpoint(ctx, podIdentityAgentEndpointIPv4)
+		cfg, err := p.getConfigWithEndpoint(ctx, podIdentityAgentEndpointIPv4)
```
simonmarty: Why was this variable renamed?

auth/auth.go
ThirdEyeSqueegee: Good idea, will do

auth/auth_test.go
ThirdEyeSqueegee: Just for consistency with the other tests

credential_provider/pod_identity_credential_provider.go
ThirdEyeSqueegee: I don't think it's needed actually, I'll drop it (klog docs [ref](https://pkg.go.dev/k8s.io/klog/v2#V))

credential_provider/pod_identity_credential_provider.go
ThirdEyeSqueegee: Just thought it was superfluous, will revert

credential_provider/pod_identity_credential_provider.go
ThirdEyeSqueegee: Consistency, we use `cfg` elsewhere in the project

simonmarty: That's just cosmetic though. Stuff like this bloats an already large PR.

server/server.go
```diff
@@ -175,18 +185,17 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
 
 	providerFactory := s.secretProviderFactory(awsConfigs, regions)
 	var fetchedSecrets []*provider.SecretValue
-	for sType := range descriptors { // Iterate over each secret type.
-		// Fetch all the secrets and update the curVerMap
-		provider := providerFactory.GetSecretProvider(sType)
-		secrets, err := provider.GetSecretValues(ctx, descriptors[sType], curVerMap)
+	for sType := range descriptors {
```
simonmarty: Why are we touching these lines?

server/server.go
```diff
@@ -148,20 +166,37 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
...
+	}
+
+	awsConfigs, err := s.getAwsConfigs(ctx, regions, usePodIdentity, preferredAddressType, roleArn, token)
 	if err != nil {
 		return nil, err
 	}
 	if len(awsConfigs) > 2 {
-		klog.Errorf("Max number of region(s) exceeded: %s", strings.Join(regions, ", "))
-		return nil, err
+		return nil, fmt.Errorf("max number of regions exceeded: %s", strings.Join(regions, ", "))
```
simonmarty: Why the change here?

server/server_test.go
```diff
@@ -909,7 +911,7 @@ var mountTests []testCase = []testCase{
 		ssmRsp:     []*ssm.GetParametersOutput{},
 		gsvRsp:     []*secretsmanager.GetSecretValueOutput{},
 		descRsp:    []*secretsmanager.DescribeSecretOutput{},
-		expErr:     "An IAM role must be associated",
+		expErr:     "IAM role must be associated",
```
simonmarty: Why the change here?

server/server.go
```diff
@@ -234,45 +283,50 @@ func (s *CSIDriverProviderServer) getAwsRegions(ctx context.Context, region, bac
...
 
 }
 
-// Private helper to get the region information for a given pod.
+// getRegionFromNode resolves the AWS region by looking up the pod's node and
+// reading the topology.kubernetes.io/region label. Falls back to AWS_REGION env var.
 //
 // When a region is not specified in the mount request, we must lookup the
-// region of the requesting pod by first descriing the pod to find the node and
+// region of the requesting pod by first describing the pod to find the node and
```
simonmarty: Ty for fixing the typos.

server/server.go
```diff
@@ -234,45 +283,50 @@ func (s *CSIDriverProviderServer) getAwsRegions(ctx context.Context, region, bac
...
-	var awsConfigsList []aws.Config
-
-	for _, region := range lookupRegionList {
-		awsAuth, err := auth.NewAuth(region, nameSpace, svcAcct, podName, preferredAddressType, eksAddonVersion, usePodIdentity, podIdentityHttpTimeout, s.k8sClient)
+// getAwsConfigs builds an AWS config for each lookup region using the pod's credentials.
+func (s *CSIDriverProviderServer) getAwsConfigs(ctx context.Context, regions []string, usePodIdentity bool, preferredAddressType, roleArn, token string) ([]aws.Config, error) {
+	var configs []aws.Config
+	appID := s.appID()
+
+	for _, region := range regions {
```
simonmarty, line 302: How big can this list of regions get? Why did we choose to iterate over them instead of making separate API calls for each?

server/server.go
ThirdEyeSqueegee: That was a bug. The `if len(awsConfigs) > 2` guard is a defensive check that fires after `getAwsConfigs` has already returned successfully (no error). The old code would return `nil, nil` which would silently succeed or cause confusing downstream behavior instead of surfacing the actual problem

server/server_test.go
ThirdEyeSqueegee: Just to match the error message in `server.go`, I'll go ahead and revert this and update the error message to match it

server/server.go
ThirdEyeSqueegee: At most 2 regions. As for iterating vs separate API calls, these aren't really API calls per se.`NewIRSACredentialProvider`/`NewPodIdentityCredentialProvider` just construct in-memory config
objects with lazy credential providers. `GetAWSConfig` calls `config.LoadDefaultConfig` which also doesn't hit the network. Credential retrieval is deferred until the first API call (`GetSecretValue` etc.)

server/server.go
simonmarty: Ok. Revert the change to the error message. Customers write regexes against those.

credential_provider/irsa_credential_provider.go
```diff
@@ -2,111 +2,46 @@ package credential_provider
...
-		&authv1.TokenRequest{
-			Spec: tokenSpec,
-		},
-		metav1.CreateOptions{})
-	if err != nil {
-		return nil, err
-	}
-
-	return []byte(tokRsp.Status.Token), nil
+func (f *csiTokenFetcher) GetIdentityToken() ([]byte, error) {
```
simonmarty, line 17: Why does this method signature support returning an error?

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -77,432 +102,134 @@ func setupMockPodIdentityAgent(t *testing.T, isIPv4, shouldFail bool) *httptest.
...
-		expError:          "",
-	}
-
-	provider, closer := newPodIdentityCredentialWithMock(t, false, testData)
-	defer func() { closer <- struct{}{} }()
-
 	cfg, err := provider.GetAWSConfig(context.Background())
-
 	if err != nil {
-		t.Fatalf("Expected no error, got: %v", err)
```
simonmarty, line 182: Remove all these style changes, this PR has enough going on as is.

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -77,432 +102,134 @@ func setupMockPodIdentityAgent(t *testing.T, isIPv4, shouldFail bool) *httptest.
...
+		t.Fatalf("unexpected error: %v", err)
 	}
 
 	creds, err := cfg.Credentials.Retrieve(context.Background())
 	if err != nil {
-		t.Fatalf("Unexpected credential retrieval error: %v", err)
+		t.Fatalf("unexpected credential retrieval error: %v", err)
```
simonmarty: Same here, this is out of scope for a feature PR that is already very complex.

credential_provider/credential_provider_test.go
```diff
@@ -1,74 +1,7 @@
...
-	"github.com/aws/aws-sdk-go-v2/service/sts"
-	"github.com/aws/aws-sdk-go-v2/service/sts/types"
-
-	authv1 "k8s.io/api/authentication/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	k8sv1 "k8s.io/client-go/kubernetes/typed/core/v1"
-)
-
 const (
 	testNamespace      = "someNamespace"
```
simonmarty: What's the point of these constants if the rest of the tests in the file are gone?

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -6,44 +6,74 @@ import (
...
+		{"ipv4", "ipv4"},
+		{"ipv6", "ipv6"},
+		{"IPv4", "ipv4"},
+		{"IPv6", "ipv6"},
+		{"invalid", "auto"},
 	}
 
 	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
+		t.Run(tt.input, func(t *testing.T) {
```
simonmarty: Why did we change this? The more stuff we change in both the application code and tests during a refactor like this one, the more likely we are to create test coverage gaps. This also makes code review harder for your reviewers.

Keep stylistic changes out of complex code reviews. This did not need to change. Fixing typos, updating comments is fine since it materially improves the quality of the code. The value add of the changes here are more dubious.

credential_provider/pod_identity_credential_provider.go
```diff
@@ -77,31 +42,25 @@ type PodIdentityCredentialProvider struct {
...
+// Callers must ensure region and token are non-empty.
 func NewPodIdentityCredentialProvider(
-	region, nameSpace, svcAcc, podName, preferredAddressType string,
+	region, preferredAddressType string,
 	podIdentityHttpTimeout *time.Duration,
-	appID string,
-	k8sClient k8sv1.CoreV1Interface,
+	appID, token string,
 ) (ConfigProvider, error) {
-	if region == "" {
```
lohdipak, line 86: Do we still need this check?

server/server.go
```diff
@@ -114,10 +130,12 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
 	failoverRegion := attrib[failoverRegionAttrib]
 	usePodIdentityStr := attrib[usePodIdentityAttrib]
 	preferredAddressType := attrib[preferredAddressTypeAttrib]
+	serviceAccountTokens := attrib[serviceAccountTokensAttrib]
 
-	// Validate preferred address type
-	if preferredAddressType != "ipv4" && preferredAddressType != "ipv6" && preferredAddressType != "auto" && preferredAddressType != "" {
```
lohdipak, line 119: Do we still need this validation?

server/server.go
```diff
@@ -148,20 +166,37 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
...
 	if usePodIdentityStr != "" {
-		var err error
 		usePodIdentity, err = strconv.ParseBool(usePodIdentityStr)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse usePodIdentity value, error: %+v", err)
 		}
 	}
 
-	awsConfigs, err := s.getAwsConfigs(ctx, nameSpace, svcAcct, s.eksAddonVersion, regions, usePodIdentity, podName, preferredAddressType, s.podIdentityHttpTimeout)
+	// Extract the specific token needed for the chosen auth method.
```
lohdipak: Code block has two clear responsibilities - extracting tokens and getting role ARN based on auth method. It could be extracted into a separate method for better readability. any thought?

credential_provider/irsa_credential_provider.go
ThirdEyeSqueegee: It's implementing the `stscreds.GetIdentityToken` interface ([ref](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/stscreds#IdentityTokenFile.GetIdentityToken))

credential_provider/credential_provider_test.go
ThirdEyeSqueegee: `testRegion` is still in use but not in many places, I'll just use `"someRegion"` directly and get rid of this file

credential_provider/pod_identity_credential_provider.go
ThirdEyeSqueegee: On `main`, `NewPodIdentityCredentialProvider` was called directly from `auth.GetAWSConfig`, where the region came from user input that could theoretically be empty. On the `tokens` branch, the call site is `server.go`'s `getAwsConfigs`, which iterates over a list of regions already validated by `getAwsRegions` (which errors if no region is found). So the empty region case is impossible by the time the credential provider is constructed.

credential_provider/irsa_credential_provider.go
simonmarty: Ah thanks, I think I've spent too much time writing Rust. I forgot Golang interfaces are implicitely implemented. And you already have a comment pointing to the relevant interface. 👍

credential_provider/pod_identity_credential_provider_test.go
```diff
@@ -60,11 +92,6 @@ func setupMockPodIdentityAgent(t *testing.T, isIPv4, shouldFail bool) *httptest.
 		Listener:    listener,
 		EnableHTTP2: true,
 		Config: &http.Server{Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if shouldFail {
```
simonmarty, line 63: Can you explain why we are comfortable getting rid of this test mode? It seems like it should drop code coverage, at least from a functional standpoint.

main.go
```diff
@@ -58,7 +57,8 @@ func parsePodIdentityHttpTimeout(timeoutStr string) *time.Duration {
 // requests.
 func main() {
 
-	klog.Infof("Starting %s version %s", auth.ProviderName, server.Version)
+	klog.Infof("Starting %s version %s", server.ProviderName, server.Version)
+	klog.Infof("This provider requires tokenRequests to be configured in the CSIDriver spec (audiences: sts.amazonaws.com, pods.eks.amazonaws.com)")
```
simonmarty, line 61: Do we have any way of conditionally emitting this log message?

ThirdEyeSqueegee: `tokenRequests` is configured on the `CSIDriver` object, and the provider doesn't know whether it's set until a mount request actually arrives. The tokens (or lack thereof) show up in the `csi.storage.k8s.io/serviceAccount.tokens` attribute of each `MountRequest`. `server.go` L138 calls `utils.ParseServiceAccountTokens(serviceAccountTokens)` which returns an error if the tokens attribute is empty/missing, with the message "serviceAccount.tokens not provided - ensure tokenRequests is configured in CSIDriver"

credential_provider/pod_identity_credential_provider_test.go
ThirdEyeSqueegee: On `main`, `shouldFail` was used by tests like `TestPodIdentityCredentialProvider_IPv4Failure` that verified behavior when the Pod Identity Agent returned a 500. Those tests worked by passing `shouldFail: true` to the mock server and then asserting "failed to refresh cached credentials". We rewrote these tests; instead of a mock server that returns 500, the tests point the endpoint at an unreachable address (http://127.0.0.1:1) which is a more realistic failure mode (agent not reachable) and doesn't need the `shouldFail` parameter (see `TestPodIdentityCredentialProvider_GetAWSConfig_BothFail`)

simonmarty: Those are testing different things, there are situations where the pod identity agent will return 500s despite being reachable. https://github.com/aws/eks-pod-identity-agent/blob/main/pkg/handlers/eks_credential_handler.go#L82-L95

If you want to add this test case incrementally, go for it. I think it's a valuable test scenario. Avoid changing test cases at the same time that you're doing a major refactor.

server/server.go
ThirdEyeSqueegee: That validation was replaced by `parseAddressPreference` in `pod_identity_credential_provider.go`, which handles invalid values by defaulting to `"auto"` instead of erroring out. On `main`, an invalid `preferredAddressType` like `"foobar"` would reject the entire mount request. On `tokens`, it falls back to auto. `parseAddressPreference` already existed on `main` with the same fallback-to-auto behavior. The server-level validation is redundant and stricter than necessary since it would reject values that the credential provider would have handled fine.

server/server.go
ThirdEyeSqueegee: Yep good call, I'll push another commit that factors it out

ThirdEyeSqueegee: Done :D

credential_provider/pod_identity_credential_provider_test.go
ThirdEyeSqueegee: Covered in the latest commit 🫡

server/server.go
```diff
@@ -239,31 +295,39 @@ func (s *CSIDriverProviderServer) getAwsRegions(ctx context.Context, region, bac
...
+
+		if usePodIdentity {
+			credProvider, err = credential_provider.NewPodIdentityCredentialProvider(region, preferredAddressType, s.podIdentityHttpTimeout, appID, token)
+		} else {
+			credProvider, err = credential_provider.NewIRSACredentialProvider(region, roleArn, appID, token)
+		}
 		if err != nil {
 			return nil, fmt.Errorf("%s: %s", region, err)
 		}
-		awsConfig, err := awsAuth.GetAWSConfig(ctx)
```
simonmarty, line 251: Why was this renamed?

server/server_test.go
```diff
@@ -1373,7 +1375,7 @@ var mountTestsForMultiRegion []testCase = []testCase{
 			{"objectName": "TestSecret1", "objectType": "secretsmanager"},
 			{"objectName": "TestParm1", "objectType": "ssmparameter"},
 		},
-		expErr:     "fakeRegion: An IAM role must be associated",
+		expErr:     "An IAM role must be associated",
```
simonmarty, line 1378: Why was the expected error changed?

simonmarty: Discussed offline, we no longer have the region in the context since this is happening beforehand.

ThirdEyeSqueegee: `getRoleArn` is no longer being called from `getAwsConfigs`, which is where `fakeRegion` was being tacked on, but rather in `resolveCredentials`

server/server.go
ThirdEyeSqueegee: Missed this in the last diff-minimizing pass, my bad. Pushing a commit to revert

PR #593, opened 2026-03-27
Add a basic benchmark

simonmarty:

1. We don't have a way to detect performance regressions in the CSI driver.

### What is changing?

1. Added a really basic set of benchmarks using Go's built-in benchmark tooling and a little github action to compare the benchmark results to the baseline.
2. Warning: The GitHub action workflow was almost entirely written by AI. I'm still investigating whether the way it's outputting benchmark diffs is the best way to do things.

### Related Links
- **Issue #, if available**:

---

### How was this tested?

1. Tested on my fork.

1. 

---

## Reviewee Checklist

  *If not, why:*
  *If not, why:*
  *If not, why:*
  *If not, why:*
  *If not, why:* Benchmark doesn't run over the network, no changes to integ tests.
- [ ] I have updated integration tests (if needed)
  *If not, why:*
  *If not, why:*
  *If not, why:*
- [ ] I have updated README/documentation (if needed)
  *If not, why:*
- [ ] I have clearly called out breaking changes (if any)
  *If not, why:* No breaking changes, this is purely incremental.

---

## Reviewer Checklist

---


.github/workflows/bench.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: actions/cache/restore@v4
+        with:
+          path: bench-baseline.txt
+          key: benchmark-baseline-${{ runner.os }}-main
+
+      - name: Compare benchmarks
+        if: github.event_name == 'pull_request' && hashFiles('bench-baseline.txt') != ''
+        run: |
+          go install golang.org/x/perf/cmd/benchstat@latest
+          benchstat bench-baseline.txt bench-result.txt | tee bench-comparison.txt
```
simonmarty: From what I'm reading online this seems to be the right way to do benchmark comparisons. I'm not a huge fan of the custom script below to publish the benchmark comparison results :/

.github/workflows/bench.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Download previous benchmark result
+        if: github.event_name == 'pull_request'
+        uses: actions/cache/restore@v4
+        with:
+          path: bench-baseline.txt
+          key: benchmark-baseline-${{ runner.os }}-main
+
+      - name: Compare benchmarks
+        if: github.event_name == 'pull_request' && hashFiles('bench-baseline.txt') != ''
```
reyhankoyun: `hashFiles()` evaluates against the workspace at checkout time, before the cache restore on line 30 runs. This will always return `''`, so the comparison step will never execute. Maybe we could replace with a file check inside the run:
```
  - name: Compare benchmarks                                                    
     if: github.event_name == 'pull_request'                                     
     run: |                                                                      
         if [ -f bench-baseline.txt ]; then                                        
            go install golang.org/x/perf/cmd/benchstat@latest                       
            benchstat bench-baseline.txt bench-result.txt | tee bench-comparison.txt
        fi
```

.github/workflows/bench.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: |
+          go install golang.org/x/perf/cmd/benchstat@latest
+          benchstat bench-baseline.txt bench-result.txt | tee bench-comparison.txt
+
+      - name: Post benchmark comparison to PR
+        if: github.event_name == 'pull_request' && hashFiles('bench-comparison.txt') != ''
```
reyhankoyun: Same issue as above

.github/workflows/bench.yml
reyhankoyun: One option is [benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) which handles PR commenting, threshold alerts, `fail-on-alert`, and history charts out of the box with native Go support. It would replace lines 42-68 and the cache/benchstat steps. However, it does simple percentage comparison rather than `benchstat`'s statistical analysis, so you'd lose the confidence intervals from `-count=5`. Depends on whether you value the statistical rigor or the CI features more.

server/benchmark_test.go
```diff
@@ -0,0 +1,155 @@
...
+		},
+		descRsp: []*secretsmanager.DescribeSecretOutput{},
+		perms:   "420",
+	}
+
+	for b.Loop() {
+		dir := b.TempDir()
+		svr := newServerWithMocks(&tst, true, nil)
+		req := buildBenchMountReq(dir, tst)
+		svr.Mount(context.Background(), req)
```
reyhankoyun: The error return from `Mount` is discarded in all four benchmarks. If the mock setup is wrong, you're silently benchmarking error paths. You could add:                  
```                                                                   
if _, err := svr.Mount(context.Background(), req); err != nil {               
    b.Fatal(err)                                                              
}                                                                             
```

server/benchmark_test.go
```diff
@@ -0,0 +1,155 @@
...
+		},
+		ssmRsp: []*ssm.GetParametersOutput{},
+		gsvRsp: []*secretsmanager.GetSecretValueOutput{
+			{SecretString: aws.String("secret1"), VersionId: aws.String("1")},
+		},
+		descRsp: []*secretsmanager.DescribeSecretOutput{},
+		perms:   "420",
+	}
+
+	for b.Loop() {
```
reyhankoyun, line 65: nit: Since setup (`newServerWithMocks` + `b.TempDir()`) runs inside the loop, the benchmark measures end-to-end including mock construction and filesystem ops. Might be worth a brief comment noting that's intentional, so future readers don't try to "fix" it.

.github/workflows/bench.yml
simonmarty: I'll give that action a shot.

server/benchmark_test.go
simonmarty: I can add a comment explaining what the purpose of the benchmark is.

server/benchmark_test.go
simonmarty: Nice, fixing.

.github/workflows/bench.yml
simonmarty: Done in 5faa261260ef2f768db43b61e8af4b3d40a95d58

Got a successful baseline in the main branch workflow pushed https://github.com/simonmarty/secrets-store-csi-driver-provider-aws/actions/runs/24360933652/job/71140187683

And a successful comparison in a test PR on my fork https://github.com/simonmarty/secrets-store-csi-driver-provider-aws/actions/runs/24361094607

simonmarty:
Latest successful benchmark comparison here: https://github.com/simonmarty/secrets-store-csi-driver-provider-aws/actions/runs/24364877168

server/benchmark_test.go
simonmarty: Done in https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/593/changes/d2a507d0ec6b5490ab95df8d8f83d2f388ae556b

server/benchmark_test.go
simonmarty: Done in https://github.com/aws/secrets-store-csi-driver-provider-aws/pull/593/changes/3d0669d05c5ddc13aa7add6c2810be0a8356ae12

server/benchmark_test.go
```diff
@@ -0,0 +1,171 @@
...
+		gsvRsp:  []*secretsmanager.GetSecretValueOutput{},
+		descRsp: []*secretsmanager.DescribeSecretOutput{},
+		perms:   "420",
+	}
+
+	for b.Loop() {
+		dir := b.TempDir()
+		svr := newServerWithMocks(&tst, true, nil)
+		req := buildBenchMountReq(dir, tst)
+		svr.Mount(context.Background(), req)
```
reyhankoyun: This one still has the stray unchecked Mount call. We can remove it for consistency with the error checking elsewhere.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Compare benchmarks
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          tool: 'go'
+          output-file-path: bench-result.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          external-data-json-path: bench-baseline.json
+          comment-on-alert: true
+          summary-always: true
```
reyhankoyun, line 44: Without `fail-on-alert: true` or `alert-threshold`, a massive regression won't block the PR. I guess the question is do we want the benchmark to be informational only or do we want it to block regressions. Not a blocker, just curious.

simonmarty: I was leaning towards informational, at least initially. We can make it fail later once we're confident in the data it provides. Otherwise we don't have permissions to override if it causes false positives.

server/benchmark_test.go
simonmarty: Sorry, I missed it in my find and replace.

.github/workflows/bench.yml
reyhankoyun: Sounds good to me

reyhankoyun (approved):

.github/workflows/bench.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Run benchmarks
+        run: go test -bench=. -benchmem -count=5 -timeout=30m ./server/ | tee bench-result.txt
+
+      - name: Download previous benchmark result
+        if: github.event_name == 'pull_request'
+        uses: actions/cache/restore@v5
+        with:
+          path: bench-baseline.json
+          key: benchmark-baseline-${{ runner.os }}-${{ github.event.pull_request.base.sha }}
```
reyhankoyun, line 34: As you pointed out on the agent, cache entries are immutable and the exact-key restore here might silently miss when the PR base SHA doesn't match a saved key. The [GitHub docs](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache) recommend using a unique-per-run key with a restore-keys prefix fallback in a single actions/cache step, which lets you drop the separate save step too. I just fixed this on the [agent repo](https://github.com/aws/aws-secretsmanager-agent/pull/191/changes#diff-22eef62c9d7b36d02f5fd5aaada92702e22d6795c464239cb7b8fe0f26ea1e1cR31-R37)

simonmarty: The cache entry for the base sha should always exist since we run the action on every push.

reyhankoyun: I'm okay to ship as it'll work in the common case, and a miss would just mean no comparison (not a failure). The `github-action-benchmark` step will still run, it just won't have a baseline to compare against. Non-blocking.

reyhankoyun (approved):

PR #646, opened 2026-07-31
Extend traversal check to jmesPath objectAlias and mount path

harsheejshah:
Apply the `../` check to `jmesPath[].objectAlias` in addition to the top-level alias, and validate that the resolved `GetMountPath()` stays inside `GetMountDir()` before writing.

  1. `validateSecretDescriptor` applies the `../` traversal check (`badPathRE`) to the top-level `objectAlias`, but the per-entry validation loop for `jmesPath` only checks that `Path`/`ObjectAlias` are non-empty and that `filePermission` parses. When `pathTranslation` is set to `False`, a `jmesPath[].objectAlias` containing `../` sequences resolves outside
  the pod's mount directory once `getJmesEntrySecretDescriptor` copies it into a new `SecretDescriptor`. This change closes that gap and adds a general-purpose mount-path guard so future descriptor construction paths cannot bypass validation either.

  ### What is changing?

  1. Factor the existing `badPathRE` check in `provider/secret_descriptor.go` into a `validateNoPathTraversal` helper and call it for every `jmesPath` sub-entry inside `validateSecretDescriptor`.
  2. Add `ValidateMountPath` on `SecretDescriptor` (uses `filepath.Rel` against `GetMountDir()`), which asserts the resolved `GetMountPath()` stays inside the mount directory regardless of which field populated the file name.
  3. Call `ValidateMountPath` in `server.writeFile` before `os.Rename`, so any descriptor that ever reaches the write path is checked one last time.
  4. Add regression tests covering jmesPath sub-entry traversal, an end-to-end reproduction with `pathTranslation: False`, the pre-existing top-level traversal path, and `ValidateMountPath` directly.

  ### Related Links
  - **Issue #, if available**: N/A
  
  ---

  ## Testing

  ### How was this tested?

  1. New unit tests in `provider/secret_descriptor_test.go`:
     - `TestJMESPathObjectAliasTraversal` — table-driven cases covering `../` at the start, in the middle, and at the trailing edge of a `jmesPath[].objectAlias`; each must be rejected with `path can not contain ../`.
     - `TestJMESPathObjectAliasEscapesMountDirRejected` — end-to-end reproduction using the documented `pathTranslation: "False"` attribute and a deeply nested `../` alias; `NewSecretDescriptorList` must fail before any file write can happen.
     - `TestTopLevelObjectAliasTraversal` — confirms the pre-existing top-level `objectAlias` traversal check still fires (regression guard).
     - `TestValidateMountPathRejectsEscape` — table-driven direct exercise of `ValidateMountPath` for parent, deep, and embedded `../` sequences under a synthetic mount dir.
     - `TestValidateMountPathAcceptsInsideDir` — confirms normal in-directory names are still accepted.
  2. Existing `TestTraversal`, `TestNotTraversal`, and `TestGetPath` continue to pass, so the top-level behaviour and normal mount paths are unchanged.

  1. `go test ./provider/... -run 'Traversal|MountPath|JMESPath' -v`
  2. `go test ./...` — full suite passes with no regressions.

  ---
  
  ## Reviewee Checklist

    *If not, why:* Change is confined to descriptor validation and the pre-write guard in `writeFile`; no AWS API surface, IAM, or provider-plugin behaviour changed. Unit tests cover the affected code paths end-to-end.
  - [ ] I have updated integration tests (if needed)
    *If not, why:* Not needed — the new behaviour is a purely local validation reject; there is no new success path to exercise against AWS.
  - [ ] I have updated README/documentation (if needed)
    *If not, why:* The existing README does not describe per-attribute validation rules; the behaviour change (rejecting `../` in `jmesPath[].objectAlias`) matches the already-documented top-level behaviour, so no doc update is required.
    *If not, why:* No breaking changes. A `SecretProviderClass` whose `jmesPath[].objectAlias` contains `../` will now be rejected at mount time; any real-world alias (a plain filename) is unaffected.

  ---

  ## Reviewer Checklist
  
  ---

### DIFF
--- a/provider/secret_descriptor.go
+++ b/provider/secret_descriptor.go
@@ -12,9 +12,6 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-// An RE pattern to check for bad paths
-var badPathRE = regexp.MustCompile(`(\/\.\.\/)|(^\.\.\/)|(\/\.\.$)`)
-
 // An RE pattern to check for valid file permission
 var validFilePermissionRE = regexp.MustCompile("^[0-7]{4}$")
 
@@ -145,6 +142,26 @@ func (p *SecretDescriptor) GetMountPath() string {
 	return filepath.Join(p.GetMountDir(), p.GetFileName())
 }
 
+// Rejects file names that would escape or degenerate against the mount
+// directory: empty, ".", non-canonical, or non-local. Inspects
+// GetFileName() only.
+func (p *SecretDescriptor) ValidateMountPath() error {
+	name := p.GetFileName()
+	obj := p.ObjectName
+	if len(obj) == 0 {
+		obj = p.ObjectAlias
+	}
+	if name == "" {
+		return fmt.Errorf("file name is empty for object: %s", obj)
+	}
+	// Go 1.21+ considers "." local, but writing to mountDir/. would target
+	// the mount directory itself rather than a file inside it.
+	if name == "." || filepath.Clean(name) != name || !filepath.IsLocal(name) {
+		return fmt.Errorf("file name escapes mount directory: %q (object: %s)", name, obj)
+	}
+	return nil
+}
+
 // Return the object type (ssmparameter, secretsmanager, or ssm)
 func (p *SecretDescriptor) getObjectType() (otype string) {
 	oType := p.ObjectType
@@ -257,9 +274,9 @@ func (p *SecretDescriptor) validateSecretDescriptor(regions []string) error {
 		return fmt.Errorf("ssm parameters can not specify both objectVersion and objectVersionLabel: %s", p.ObjectName)
 	}
 
-	// Do not allow ../ in a path when translation is turned off
-	if badPathRE.MatchString(p.GetFileName()) {
-		return fmt.Errorf("path can not contain ../: %s", p.ObjectName)
+	// Reject file names that would escape the mount directory
+	if err := p.ValidateMountPath(); err != nil {
+		return err
 	}
 
 	// Ensure the string file permission is valid octal
@@ -284,6 +301,10 @@ func (p *SecretDescriptor) validateSecretDescriptor(regions []string) error {
 			return err
 		}
 
+		jmesDescriptor := p.getJmesEntrySecretDescriptor(&jmesPathEntry)
+		if err := jmesDescriptor.ValidateMountPath(); err != nil {
+			return err
+		}
 	}
 
 	if len(p.FailoverObject.ObjectName) > 0 {
--- a/provider/secret_descriptor_test.go
+++ b/provider/secret_descriptor_test.go
@@ -551,6 +551,9 @@ func TestTraversal(t *testing.T) {
         `, `
         - objectName: "mypath/.."
           objectType: ssmparameter
+        `, `
+        - objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:mypath"
+          objectAlias: "a/../b"
         `,
 	}
 
@@ -558,8 +561,8 @@ func TestTraversal(t *testing.T) {
 
 		_, err := NewSecretDescriptorList("/", "False", obj, singleRegion)
 
-		if err == nil || !strings.Contains(err.Error(), "path can not contain ../") {
-			t.Errorf("Expected error: path can not contain ../, got error: %v\n%v", err, obj)
+		if err == nil || !strings.Contains(err.Error(), "escapes mount directory") {
+			t.Errorf("Expected error: escapes mount directory, got error: %v\n%v", err, obj)
 		}
 
 	}
@@ -990,4 +993,147 @@ func TestValidateDescriptorFilePermission(t *testing.T) {
 			t.Errorf("got: %v want: %v", got, want)
 		}
 	})
+}
+
+// jmesPath sub-entry objectAlias containing ../ must be rejected.
+func TestJMESPathObjectAliasTraversal(t *testing.T) {
+	objects := []string{
+		`
+        - objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:jsonSecret"
+          objectAlias: "safeAlias"
+          jmesPath:
+            - path: "payload"
+              objectAlias: "../../../../etc/cron.d/pwn"
+        `, `
+        - objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:jsonSecret"
+          objectAlias: "safeAlias"
+          jmesPath:
+            - path: "payload"
+              objectAlias: "sub/../../escape"
+        `, `
+        - objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:jsonSecret"
+          objectAlias: "safeAlias"
+          jmesPath:
+            - path: "payload"
+              objectAlias: ".."
+        `,
+	}
+
+	for _, obj := range objects {
+		_, err := NewSecretDescriptorList("/mountpoint", "False", obj, singleRegion)
+		if err == nil || !strings.Contains(err.Error(), "escapes mount directory") {
+			t.Errorf("Expected error: escapes mount directory, got error: %v\n%v", err, obj)
+		}
+	}
+}
+
+// jmesPath objectAlias with ../ segments must not escape mountDir when
+// pathTranslation is disabled.
+func TestJMESPathObjectAliasEscapesMountDirRejected(t *testing.T) {
+	mountDir := "/var/lib/kubelet/pods/POD-UID/volumes/kubernetes.io~csi/secrets-store/mount"
+	objectSpec := `
+- objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:my-json-secret"
+  objectType: "secretsmanager"
+  jmesPath:
+    - path: "payload"
+      objectAlias: "../../../../../../../../../etc/cron.d/pwn"
+`
+
+	_, err := NewSecretDescriptorList(mountDir, "False", objectSpec, singleRegion)
+	if err == nil {
+		t.Fatalf("expected traversal error, but NewSecretDescriptorList accepted a jmesPath objectAlias with ../ segments")
+	}
+	if !strings.Contains(err.Error(), "escapes mount directory") {
+		t.Fatalf("expected error: escapes mount directory, got: %v", err)
+	}
+}
+
+// Confirm the pre-existing top-level objectAlias traversal check still fires.
+func TestTopLevelObjectAliasTraversal(t *testing.T) {
+	objectSpec := `
+- objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:my-secret"
+  objectAlias: "../../../../etc/cron.d/pwn"
+`
+	_, err := NewSecretDescriptorList("/mountpoint", "False", objectSpec, singleRegion)
+	if err == nil || !strings.Contains(err.Error(), "escapes mount directory") {
+		t.Fatalf("expected traversal error, got: %v", err)
+	}
+}
+
+// translate="" mirrors NewSecretDescriptorList's state when pathTranslation
+// is "False", which is when a ".." element can survive to the write path.
+// translate="/" is the only setting under which GetFileName() can return an
+// absolute name (the TrimLeft in GetFileName's else branch is skipped).
+func TestValidateMountPathRejectsEscape(t *testing.T) {
+	cases := []struct {
+		name        string
+		translate   string
+		objectAlias string
+	}{
+		{"parent-traversal", "", "../etc/passwd"},
+		{"deep-traversal", "", "../../../../etc/cron.d/pwn"},
+		{"embedded-traversal", "", "sub/../../escape"},
+		{"bare-parent", "", ".."},
+		{"trailing-parent", "", "mypath/.."},
+		{"absorbed-parent", "", "/../pathTest-abc123"},
+		{"balanced-parent", "", "a/../b"},
+		{"dot", "", "."},
+		{"empty-after-trim", "", "/"},
+		{"translate-slash-absolute", "/", "/prod/db/password"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			d := SecretDescriptor{
+				ObjectAlias: tc.objectAlias,
+				translate:   tc.translate,
+			}
+			if err := d.ValidateMountPath(); err == nil {
+				t.Fatalf("expected ValidateMountPath to reject alias %q (translate=%q), got nil",
+					tc.objectAlias, tc.translate)
+			}
+		})
+	}
+}
+
+func TestValidateMountPathDefaultTranslateAcceptsSubstituted(t *testing.T) {
+	d := SecretDescriptor{
+		ObjectAlias: "../etc/passwd",
+		mountDir:    "/mountpoint",
+		translate:   "_",
+	}
+	if err := d.ValidateMountPath(); err != nil {
+		t.Fatalf("unexpected error after slash substitution: %v (resolved to %q)",
+			err, d.GetMountPath())
+	}
+	if got, want := d.GetFileName(), ".._etc_passwd"; got != want {
+		t.Fatalf("expected resolved name %q, got %q", want, got)
+	}
+}
+
+// ValidateMountPath accepts normal in-directory file names.
+func TestValidateMountPathAcceptsInsideDir(t *testing.T) {
+	d := SecretDescriptor{
+		ObjectName: "secret",
+		ObjectType: "secretsmanager",
+		mountDir:   "/mountpoint",
+	}
+	if err := d.ValidateMountPath(); err != nil {
+		t.Fatalf("unexpected error for in-directory file: %v", err)
+	}
+}
+
+// Verifies getJmesEntrySecretDescriptor propagates translate/mountDir: under
+// default slash-substitution, a jmesPath alias with `..` segments becomes a
+// flat file name and is accepted.
+func TestJMESPathObjectAliasAcceptedUnderDefaultTranslate(t *testing.T) {
+	objectSpec := `
+- objectName: "arn:aws:secretsmanager:us-west-2:123456789012:secret:jsonSecret"
+  objectAlias: "safeAlias"
+  jmesPath:
+    - path: "payload"
+      objectAlias: "sub/../x"
+`
+	if _, err := NewSecretDescriptorList("/mountpoint", "", objectSpec, singleRegion); err != nil {
+		t.Fatalf("unexpected error under default translate: %v", err)
+	}
 }
--- a/server/server.go
+++ b/server/server.go
@@ -207,6 +207,15 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
 		fetchedSecrets = append(fetchedSecrets, secrets...) // Build up the list of all secrets
 	}
 
+	// Validate every descriptor before the first write so a rejection does
+	// not leave a partially populated mount.
+	for _, secret := range fetchedSecrets {
+		if err := secret.Descriptor.ValidateMountPath(); err != nil {
+			klog.Errorf("Refusing to write secret outside mount directory: %s", err)
+			return nil, err
+		}
+	}
+
 	// Write out the secrets to the mount point after everything is fetched.
 	var files []*v1alpha1.File
 	for _, secret := range fetchedSecrets {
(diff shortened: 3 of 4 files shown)


provider/secret_descriptor.go
```diff
@@ -145,6 +145,20 @@ func (p *SecretDescriptor) GetMountPath() string {
 	return filepath.Join(p.GetMountDir(), p.GetFileName())
 }
 
+// Validates that GetMountPath() resolves inside GetMountDir().
+func (p *SecretDescriptor) ValidateMountPath() error {
```
simonmarty, line 147: Can we canonicalize the path to an absolute path using [this method](https://pkg.go.dev/path/filepath#Abs) and just check that the path is a subpath of our secret directory? That would be simpler than the checks below.

simonmarty (dismissed):

simonmarty:
> which asserts the resolved GetMountPath() stays inside the mount directory regardless of which field populated the file name.

Let's just do this then. I don't think it matters much if there are relative paths, we just need to get the absolute path and check it is a subpath of the mount directory.

provider/secret_descriptor.go
simonmarty: Actually this method seems to be the one you actually want to use https://pkg.go.dev/path/filepath#IsLocal

harsheejshah: this should work, updating in next rev

provider/secret_descriptor.go
```diff
@@ -145,6 +145,20 @@ func (p *SecretDescriptor) GetMountPath() string {
...
+	mountPath := filepath.Clean(p.GetMountPath())
+	if mountPath == mountDir {
+		return fmt.Errorf("resolved mount path %q equals mount directory", mountPath)
+	}
+	rel, err := filepath.Rel(mountDir, mountPath)
+	if err != nil || rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+		return fmt.Errorf("resolved mount path %q escapes mount directory %q", mountPath, mountDir)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
simonmarty: The other edge case I can think of is symlinks, there might be a way to address those with filepath methods too.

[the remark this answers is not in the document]
harsheejshah: addressed

provider/secret_descriptor.go
```diff
@@ -145,6 +142,16 @@ func (p *SecretDescriptor) GetMountPath() string {
 	return filepath.Join(p.GetMountDir(), p.GetFileName())
 }
 
+// Validates that GetMountPath() resolves inside GetMountDir(). IsLocal
+// rejects ../ traversal, absolute paths, and empty names, i.e. any file name
+// that would not stay within the mount directory.
+func (p *SecretDescriptor) ValidateMountPath() error {
+	if !filepath.IsLocal(p.GetFileName()) {
```
ThirdEyeSqueegee: **Blocking, security.** `IsLocal` cleans lexically first, so balanced and trailing `..` survive. The old regex had explicit `(\/\.\.\/)` and `(\/\.\.$)` branches.

Now accepted, rejected on `main`: `mypath/..` (both objectTypes), `arn:...:secret:mypath/..`, `arn:...:secret:/../pathTest-abc123`, alias `a/../b`.

`mypath/..` makes `GetMountPath()` resolve to `GetMountDir()`. `/../pathTest-abc123` absorbs the ARN prefix, landing the secret at a filename never specified.

The driver's rule: `fileutil.Validate` rejects any `..` element (`validatePath`, `atomic_writer.go:279`) and requires `filepath.Clean(p) == p` (`writer.go:34`). Something along these lines would match it:

```go
name := p.GetFileName()
if name == "" || filepath.Clean(name) != name || !filepath.IsLocal(name) {
    return fmt.Errorf(...)
}
```

ThirdEyeSqueegee: **Compatibility, breaking change.** `pathTranslation` accepts any single character, and `/` is legal. With `translate == "/"`, `GetFileName()` takes the substitution branch and skips the leading-slash `TrimLeft`, leaving the name absolute.

objectName `/prod/db/password` (the common SSM shape) is accepted on `main` (mount path `/mnt/prod/db/password`) and rejected here with `escapes mount directory`. `a/../b` flips the other way. The checklist marks "no breaking changes". No test in this PR passes a non-empty, non-default `translate`.

ThirdEyeSqueegee: **Test coverage.** The check runs on the post-translation name (`GetFileName()` translates at 122-126), which makes it near-vacuous under the default `_`: alias `../etc/passwd` becomes `.._etc_passwd`, `IsLocal` true, accepted.

Every traversal test in this PR passes `"False"`, so the default configuration is untested. A default-translate case would help, along with a decision on whether substitution is intended as a sanitizer.

provider/secret_descriptor.go
```diff
@@ -145,6 +142,16 @@ func (p *SecretDescriptor) GetMountPath() string {
...
+		return fmt.Errorf("resolved mount path %q escapes mount directory %q", p.GetMountPath(), p.GetMountDir())
```
ThirdEyeSqueegee: **Diagnosability.** Three problems:

- Drops `p.ObjectName`. Every other error in `validateSecretDescriptor` includes it; with 20 objects mounted an operator has to bisect. `GetMountPath()` has already `Join`-ed the offending name away, so no input is named.
- Self-contradictory for empty names. objectName `/` with `pathTranslation: False` gives `resolved mount path "/mountpoint" escapes mount directory "/mountpoint"` — identical strings.
- Embeds the kubelet mount path including the pod UID; returned over gRPC and klog'd (`server.go:193`).

Suggested: `fmt.Errorf("file name escapes mount directory: %s", p.ObjectName)`.

provider/secret_descriptor.go
```diff
@@ -145,6 +142,16 @@ func (p *SecretDescriptor) GetMountPath() string {
 	return filepath.Join(p.GetMountDir(), p.GetFileName())
 }
 
+// Validates that GetMountPath() resolves inside GetMountDir(). IsLocal
```
ThirdEyeSqueegee: **Documentation.** Three claims the code does not provide:

- "resolves inside `GetMountDir()`" — false for `mypath/..`, which resolves *to* the mount dir, and for symlinked intermediate components (`IsLocal` is lexical; `linkdir/victim` passes with `linkdir -> outside`).
- "rejects absolute paths" — unreachable. `GetFileName()` either `TrimLeft`s every leading separator or substitutes it, so no `translate` value yields an absolute name. The `TrimLeft` is the actual protection, and this comment invites removing it as redundant.
- "rejects empty names" — only when `translate` is empty. Under the default `_`, objectName `/` becomes `_` and is accepted.

provider/secret_descriptor_test.go
```diff
@@ -527,39 +527,28 @@ func TestGetPath(t *testing.T) {
 func TestTraversal(t *testing.T) {
```
ThirdEyeSqueegee, line 527: **Blocking, test coverage.** Each deleted row is a case the new predicate accepts:

- `arn:...:secret:/../pathTest-abc123`
- `arn:...:secret:mypath/..`
- `mypath/..`, `objectType: secretsmanager`
- `mypath/..`, `objectType: ssmparameter`

Re-added verbatim on this branch, all four are accepted. All four pass on `main`. CI is green because the failing rows were removed, and the description states the opposite.

provider/secret_descriptor_test.go
```diff
@@ -991,3 +980,103 @@ func TestValidateDescriptorFilePermission(t *testing.T) {
...
+`
+	_, err := NewSecretDescriptorList("/mountpoint", "False", objectSpec, singleRegion)
+	if err == nil || !strings.Contains(err.Error(), "escapes mount directory") {
+		t.Fatalf("expected traversal error, got: %v", err)
+	}
+}
+
+// ValidateMountPath must reject file names that would escape the mount
+// directory. translate="" mirrors the state NewSecretDescriptorList sets when
+// pathTranslation is "False"; slash-substitution translation already
```
ThirdEyeSqueegee: **Inaccurate rationale.** "slash-substitution translation already neutralizes `../` on its own" is false. Substitution replaces the separator, not the `..` segments. `../etc/passwd` under `translate="_"` becomes `.._etc_passwd`, inert because the separator is gone, not because traversal was neutralized. This claim is what justifies skipping the default-translation coverage noted on `ValidateMountPath`.

server/server.go
```diff
@@ -413,8 +418,13 @@ func (s *CSIDriverProviderServer) writeFile(secret *provider.SecretValue, mode o
 		return nil, err
 	}
 
+	mountPath := secret.Descriptor.GetMountPath()
+	if fi, err := os.Lstat(mountPath); err == nil && fi.Mode()&os.ModeSymlink != 0 {
```
ThirdEyeSqueegee: **Blocking, availability.** This block should come out. Three defects:

**Wedges mounts.** AtomicWriter makes every user-visible file a symlink (`name -> ..data/name`, `createUserVisibleFiles`, `atomic_writer.go:435`; documented README:359). Against that layout `writeFile` returns `refusing to overwrite existing symlink` instead of rotating. `Mount()` aborts on the first `writeFile` error, so one symlink fails every secret in the SecretProviderClass, every reconcile, indefinitely, pod pinned to the stale value. Nothing unlinks it. Reachable by flipping `driverWritesSecrets` true->false or downgrading the driver. Upstream `cleanupProviderFiles` (`writer.go:95`) skips symlinks deliberately.

**Guards nothing.** `os.Rename` replaces a destination symlink atomically without following it. Removing only this block and re-running `TestWriteFileRejectsPreExistingSymlink`: returns nil, sentinel still reads `original`, target becomes a regular file with the new content. The assertion at `server_test.go:3115` cannot fire on any path.

**Fails open.** `err == nil &&` discards every `Lstat` error. Inside a `0000` directory `Lstat` returns permission denied, the condition is false, the check is skipped, no log. It should distinguish `os.IsNotExist` (proceed) from other errors (surface). The inline `err` also shadows the outer `err`; `go vet` does not flag it.

There's also a check-then-use race between this line and the `Rename` at 427.

server/server_test.go
```diff
@@ -2989,6 +2989,168 @@ func TestMountMaxRegionsExceeded(t *testing.T) {
 	}
 }
 
+// writeFile must reject a SecretDescriptor whose file name would escape the
+// mount directory. Guards descriptors that reach the write path through any
+// construction path (including ones that skip NewSecretDescriptorList).
+func TestWriteFileRejectsTraversal(t *testing.T) {
```
ThirdEyeSqueegee, line 2994: **Test coverage.** `TestWriteFileRejectsTraversal` passes with the `ValidateMountPath` block deleted from `writeFile`: `os.CreateTemp` independently fails with `pattern contains path separator` for any name containing `/`. Only `TestWriteFileRejectsTraversalInDriverWriteMode` kills that mutation.

Lines 1038 and 1064 have the same defect: `(contains path separator)|(escapes mount directory)` passes whether the guard fired or `CreateTemp` failed incidentally. Asserting on `escapes mount directory` alone, or using a separator-free traversal name (`..`), would fix both.

server/server_test.go
```diff
@@ -2989,6 +2989,168 @@ func TestMountMaxRegionsExceeded(t *testing.T) {
...
+	}
+
+	svr := &CSIDriverProviderServer{}
+	if _, err := svr.writeFile(secret, 0644); err == nil {
+		t.Fatalf("expected writeFile to reject a traversing descriptor, got nil")
+	}
+
+	// Confirm no file was written outside the mount dir.
+	if _, err := os.Stat("/etc/cron.d/pwn"); err == nil {
+		t.Fatalf("writeFile created /etc/cron.d/pwn — path traversal was not prevented")
```
ThirdEyeSqueegee: **Test quality.** `os.Stat("/etc/cron.d/pwn")` asserts on host state `writeFile` cannot produce — it only writes under `mountDir`. If the file exists for any unrelated reason the test fails with `writeFile created /etc/cron.d/pwn`. Reachability also depends on `TMPDIR` depth: on macOS the traversal resolves to `/var/folders/16/etc/...`. Asserting the mount directory is empty would be both correct and hermetic.

Separately, only this test checks `d == nil`. Lines 3058, 3103, and 3142 dereference `d` directly, so an empty descriptor map panics instead of `t.Fatalf`-ing — and a panic aborts the binary, skipping every later test in the package.

ThirdEyeSqueegee (changes requested):
`filepath.IsLocal` is weaker than the `badPathRE` it replaces, and the four deleted `TestTraversal` rows are exactly the four cases the new predicate accepts. Verified on both branches:

| objectName / alias | `main` | this PR |
|---|---|---|
| `mypath/..` (both objectTypes) | reject | accept |
| `arn:...:secret:mypath/..` | reject | accept |
| `arn:...:secret:/../pathTest-abc123` | reject | accept |
| objectAlias `a/../b` | reject | accept |

`IsLocal` cleans lexically before deciding, so balanced and trailing `..` pass. `mypath/..` resolves `GetMountPath()` to the mount directory itself.

## Asks:

1. We should restore the four deleted `TestTraversal` rows, then adjust the implementation to pass them. The predicate should match the driver's rule: rejecting any `..` path element and requiring `filepath.Clean(name) == name`. The provider is currently laxer than the driver's own `File.Path` validator, so `driverWriteSecrets` mounts pass here and fail inside the driver.
2. We should remove the `os.Lstat` symlink check. `os.Rename` never follows a destination symlink, so `main` is not vulnerable to what it guards. It hard-fails on the driver's AtomicWriter layout, where every visible file is a symlink. Not mentioned in the description.
3. The description needs correcting. It cites `filepath.Rel` and a `validateNoPathTraversal` helper, neither of which is in the diff, and states existing `TestTraversal` cases continue to pass while removing four of them.

Not caused by this PR: `TestMounts` fails identically on `main` (pre-existing, mock-ordering). Branch builds clean, `go vet` passes.

`os.OpenRoot` (go.mod is `go 1.26.0`) enforces containment per path component including symlinked intermediates, which lexical checks cannot.

---

## Three findings outside the diff

These land on unchanged lines, so they cannot be inline comments.

### `provider/secret_descriptor.go:477` — blocking, security

`seenAliases` keys on the raw alias string; the file is written at the cleaned path. The newly-permitted balanced-`..` form is therefore a cross-descriptor overwrite primitive.

`objectAlias: important` plus `objectAlias: x/../important` both parse with `err == nil` and report the same `GetMountPath()`. End-to-end under default `pathTranslation`: one file remains, holding the second secret. The first is gone, no error, no log. Same for objectName `arn:...:secret:/../pathTest-abc123` alongside alias `pathTest-abc123`.

`badPathRE` blocked this on `main`. Anyone who can add one object to a SecretProviderClass can now target another object's file.

### `provider/secrets_manager_provider.go:251` — security

The write path is guarded against symlinks, the read path is not, and `os.ReadFile` follows them.

With `mountDir/okAlias -> outside/hostfile`, `ValidateMountPath()` returns nil and `reloadSecret` (taken when `isCurrent()` is true, line 122) returns the host file contents as the secret value, into `fetchedSecrets` and, in `driverWriteSecrets` mode, back to the driver as file `Contents`.

Either the symlink is untrusted for both operations or for neither. `os.OpenRoot` would close both.

### `provider/secret_descriptor_fuzz_test.go:39` — test coverage

The target discards both return values, so it catches only panics. Adding the invariant (cleaned `GetMountPath()` strictly under cleaned `GetMountDir()`) fails on the corpus already in the repo: seed `secret:mypath/..` yields `mountPath == mountDir`. The hole is reachable from existing seeds, and CI is green only because nothing is asserted. One `strings.HasPrefix` here would have caught the `IsLocal` finding automatically.

provider/secret_descriptor.go
harsheejshah: One correction on framing: none of these actually escape the mount. mypath/.. cleans to . and Rename fails on the mount dir itself. /../pathTest-abc123 doesn't absorb the ARN prefix, GetFileName operates on ObjectAlias or 
ObjectName-as-filename, not the full ARN. It's a strictness regression from the deleted regex, worth restoring, not an escape.

harsheejshah: translate="/" is `strings.ReplaceAll(name, "/", "/")`, a literal no-op. Not a realistic deployment. Happy to add a doc note but not carving a special case. For the test ask sure can add it

provider/secret_descriptor.go
harsheejshah: Under `translate="_"` there are no separators left after substitution, so `..` can't form a path element. That's why the check looks near-vacuous, not a bug.

provider/secret_descriptor_test.go
harsheejshah: sure can add the tests

provider/secret_descriptor_test.go
harsheejshah: Substitution eliminates the separator so .. can't form a path element, not "neutralizes ../". Will update the comment and add default-translate rows to TestTraversal.

server/server.go
harsheejshah: Pushing back on the AtomicWriter framing though. AtomicWriter is driver-side. When driverWritesSecrets=false the provider writes to the target path and never sees driver-side symlinks. When driverWritesSecrets=true writeFile short-circuits before this block. The block is dead code in both modes, that's the reason to remove it.

harsheejshah: Intermediate-symlink hardening via os.OpenRoot will be owned by the client side team.

harsheejshah: Removing the block and the test.

harsheejshah:
On framing, none of the four "accepted, main rejects" cases actually escape the mount subtree. After `filepath.Clean`, `mypath/..` is the mount dir itself, `a/../b` is `b`, `/../pathTest-abc123` is inside the mount. This is a strictness regression from the deleted regex, worth fixing to match the driver's rule, not the same threat class as the objectAlias-traversal. Labeling the four as "blocking, security" reads high.
  
 On scope, this PR was opened for a specific Sauron item (jmesPath objectAlias traversal). Findings in `secrets_manager_provider.go`, `secret_descriptor.go:477`, and the fuzz test are unrelated to that item. `seenAliases` collision is real but distinct dedup logic; tightening the predicate closes the traversal path as a side effect. `reloadSecret` symlink follow is inside a per-pod tmpfs where the pod is both writer and reader, no provider privilege boundary crossed. Filing separately, not blocking this PR on them.
  
On the `os.Lstat` block, agreed it goes. Correcting the AtomicWriter framing though, AtomicWriter is driver-side, the provider writes directly when `driverWritesSecrets=false` and short-circuits before the Lstat when
  `driverWritesSecrets=true`. The block is dead code in both modes, that's why it goes, not "wedges mounts in the wild."
 
Test note: `TestMounts/Fail_Pod_Retrieval`, `TestMounts/Mount_Json_Success-specify_ARN`, and four `TestMountsNoWrite` subtests fail on `origin/main` today, unchanged by this PR. Only the PR-owned tests are relevant to CI signal here.

ThirdEyeSqueegee (dismissed):

lohdipak (dismissed):

server/server_test.go
```diff
@@ -2989,6 +2989,133 @@ func TestMountMaxRegionsExceeded(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to read mount dir: %v", err)
+	}
+	if len(entries) != 0 {
+		names := make([]string, 0, len(entries))
+		for _, e := range entries {
+			names = append(names, e.Name())
+		}
+		t.Fatalf("expected empty mount dir after rejected write, got: %v", names)
+	}
```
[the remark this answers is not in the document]
harsheejshah: addressed

ThirdEyeSqueegee (approved):

provider/secret_descriptor.go
```diff
@@ -145,6 +142,26 @@ func (p *SecretDescriptor) GetMountPath() string {
 	return filepath.Join(p.GetMountDir(), p.GetFileName())
 }
 
+// Rejects file names that would escape or degenerate against the mount
+// directory: empty, ".", non-canonical, or non-local. Inspects
+// GetFileName() only.
+func (p *SecretDescriptor) ValidateMountPath() error {
```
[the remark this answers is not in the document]
harsheejshah: fixed

server/server.go
```diff
@@ -207,6 +207,15 @@ func (s *CSIDriverProviderServer) Mount(ctx context.Context, req *v1alpha1.Mount
...
 	}
 
+	// Validate every descriptor before the first write so a rejection does
+	// not leave a partially populated mount.
+	for _, secret := range fetchedSecrets {
+		if err := secret.Descriptor.ValidateMountPath(); err != nil {
+			klog.Errorf("Refusing to write secret outside mount directory: %s", err)
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
harsheejshah: updated comment

server/server_test.go
```diff
@@ -2989,6 +2989,133 @@ func TestMountMaxRegionsExceeded(t *testing.T) {
...
+	var d *provider.SecretDescriptor
+	for _, list := range descriptors {
+		for _, entry := range list {
+			d = entry
+		}
+	}
+	if d == nil {
+		t.Fatalf("no descriptor produced")
+	}
```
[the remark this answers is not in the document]
harsheejshah: addressed

lohdipak (approved):

PR #656, opened 2026-08-24
fix: default GOPROXY to module proxy in Docker build

ThirdEyeSqueegee:

1. The Docker build pinned `GOPROXY=direct`, so `go mod download` resolved every module from its origin instead of the module proxy. That made the build depend on gopkg.in's `?go-get=1` redirector, which is returning `502 Proxy Error` for GitHub Actions runners:

   ```
   # get https://gopkg.in/yaml.v2?go-get=1: 502 Proxy Error (10.138s)
   go: gopkg.in/evanphx/json-patch.v4@v4.13.0: unrecognized import path "gopkg.in/evanphx/json-patch.v4": reading https://gopkg.in/evanphx/json-patch.v4?go-get=1: 502 Proxy Error
   ```

   This blocked the "Build and push test docker image" workflow across all four attempts of run 32774071533, while the Go workflow succeeded on the same commit because it uses the default `GOPROXY`.

### What is changing?

1. `Dockerfile`: `GOPROXY` becomes a build arg defaulting to Go's normal `https://proxy.golang.org,direct` chain, so module fetches hit the proxy cache and no longer depend on gopkg.in being reachable.
2. `Makefile`: `docker-buildx` passes the host's `go env GOPROXY` through as `--build-arg GOPROXY`, so networks that cannot reach `proxy.golang.org` keep working without editing the Dockerfile. Note that Go falls back from the proxy to `direct` only on HTTP 404/410, not on network errors, so the override is still needed on such networks.

### Related Links
- **Issue #, if available**: n/a

---

### How was this tested?

1. Both matrix legs of "Build and push test docker image" pass on this branch (run 32778267134): `ubuntu-24.04-arm` in 1m19s and `ubuntu-latest` in 1m46s. The same arm leg failed at `go mod download` on all four attempts of run 32774071533 before this change.
2. Go workflow passes on this branch (run 32778170893), including unit tests, fuzz targets, staticcheck, and Codecov.
3. Local `docker build` with the override, confirming the previous behavior is still available.

1. Build with the override, and the resulting `GOPROXY` inside the builder stage:

   ```
   $ docker build --no-cache --target go --build-arg GOPROXY=direct .
   ...
   #15 DONE 14.0s
   $ docker run --rm --entrypoint sh <image> -c 'go env GOPROXY'
   direct
   ```

2. Makefile passthrough resolving from the host:

   ```
   $ make -n docker-buildx | grep -o 'build-arg GOPROXY=[^ ]*'
   build-arg GOPROXY=direct
   ```

---

## Reviewee Checklist

  *If not, why:*
  *If not, why:*
  *If not, why:*
  *If not, why:*
  *If not, why:* Not run locally; they are running in CI on this PR via the `safe-to-test` label. This change alters only how modules are fetched during the Docker build, not provider behavior.
- [x] I have updated integration tests (if needed)
  *If not, why:* No test change needed; build-time dependency resolution only.
  *If not, why:* `ENV GOPROXY` lives only in the discarded `go` builder stage; the published image is built `FROM scratch`.
  *If not, why:*
- [x] I have updated README/documentation (if needed)
  *If not, why:* No docs reference the Docker build args or `GOPROXY`.
  *If not, why:* None. Default behavior only improves proxy usage; the previous `direct` behavior remains available via build arg.

---

## Reviewer Checklist

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ docker-buildx:
 	docker buildx build \
 				--platform linux/arm64,linux/amd64 \
 				--build-arg LDFLAGS=$(LDFLAGS) \
-				$(if $(GOPROXY),--build-arg GOPROXY=$(GOPROXY)) \
+				--build-arg 'GOPROXY=$(GOPROXY)' \
 				--push \
 				-t $(REGISTRY_NAME):latest \
 				-t $(REGISTRY_NAME):$(FULL_REV) \
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ docker-buildx:
 	docker buildx build \
 				--platform linux/arm64,linux/amd64 \
 				--build-arg LDFLAGS=$(LDFLAGS) \
+				$(if $(GOPROXY),--build-arg GOPROXY=$(GOPROXY)) \
 				--push \
 				-t $(REGISTRY_NAME):latest \
 				-t $(REGISTRY_NAME):$(FULL_REV) \
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ docker-buildx:
 	docker buildx build \
 				--platform linux/arm64,linux/amd64 \
 				--build-arg LDFLAGS=$(LDFLAGS) \
-				--build-arg 'GOPROXY=$(GOPROXY)' \
+				$(if $(GOPROXY),--build-arg 'GOPROXY=$(GOPROXY)') \
 				--push \
 				-t $(REGISTRY_NAME):latest \
 				-t $(REGISTRY_NAME):$(FULL_REV) \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,6 +19,8 @@ LDFLAGS?="-X github.com/aws/secrets-store-csi-driver-provider-aws/server.Version
 
 CHART_RELEASER_PATH ?= cr
 
+GOPROXY ?= $(shell go env GOPROXY 2>/dev/null || echo https://proxy.golang.org,direct)
```
simonmarty: Can we just leave GOPROXY unset if the env var is not present instead? (and use the Go default value).

simonmarty (changes requested):

simonmarty (approved):

bob2681312 (approved):

PR #667, opened 2026-09-14
fix: report object versions from the region that supplied the value

ThirdEyeSqueegee:

1. With `failoverRegion` set, the version reported to the driver comes from the failover copy rather than the region that supplied the mounted value. Every client runs even after the primary succeeds, and each one writes the version map keyed by file name, so the failover client's version lands last. Without a `failoverObject` the failover region resolves the same object name, and when that copy is not rotated the reported version never changes.
2. An unchanged version makes the driver's `SecretProviderClassPodStatus` update a no-op, so no watch event fires and a synced Kubernetes Secret is only rewritten on the reconciler's own timer: `RequeueAfter: 5 * time.Minute` in secrets-store-csi-driver v1.6.0. Rotation therefore takes up to five minutes to reach synced Secrets.
3. The same overwrite defeats the `DescribeSecret` version check, so every republish re-fetches from both regions instead of reloading the mounted value from disk.
4. This is also the root cause of the recent integration test flakiness on test 11 (`jmesPath for Secrets Manager secret with rotation`): its synced-Secret poll allows about 270 seconds of wall clock, while the Secret converges 295-305 seconds after rotation.

### What is changing?

1. In both providers, only the client whose values are mounted writes versions. A later client receives a copy of the map, so it can still decide whether to fetch, and its writes are discarded.
2. Failover fetching, error handling, and the 4XX fail-fast behavior are unchanged. No provider, chart, or integration test change is needed.

### Related Links
- **Issue #, if available**: none
- Driver requeue timer: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/v1.6.0/controllers/secretproviderclasspodstatus_controller.go#L358

---

### How was this tested?

1. New unit test `TestMultiRegionReportsPrimaryVersion` mounts one Secrets Manager secret and one parameter with both regions healthy and a different version in each region, then asserts that the mounted values and the reported object versions both come from the primary region.
2. Full suite on the Go version in `go.mod` (1.26.0), plus `gofmt -s -l .`, `goimports -l .`, and `staticcheck ./...`.
3. Confirmed the new test fails without the provider change, by stashing only `provider/` and re-running it.

1. New test against this change, and against the same tree with `provider/` reverted:

```
--- PASS: TestMultiRegionReportsPrimaryVersion (0.07s)
ok  	github.com/aws/secrets-store-csi-driver-provider-aws/server	0.592s

server_version_report_test.go:69: Expected TestSecret1 version primaryVersion got failoverVersion
server_version_report_test.go:69: Expected TestParm1 version 7 got 1
--- FAIL: TestMultiRegionReportsPrimaryVersion (0.06s)
```

2. Lint and analysis on this branch: `gofmt -s -l .` and `goimports -l .` print nothing, `staticcheck ./...` exits 0.

---

## Reviewee Checklist

  *If not, why:*
  *If not, why:*
  *If not, why:*
- [ ] Unit test coverage check is passing
  *If not, why:* `codecov/patch` reports 0.00% of the diff hit. CI runs `go test ./... -coverprofile` without `-coverpkg`, so coverage is attributed per package under test only, and the new test lives in the `server` package while the changed lines live in `provider`. The `provider` package has no existing unit tests for `fetchSecretManagerValue` or `fetchParameterStoreValue`. Happy to add a provider-package test that constructs fake clients directly if the coverage gate should go green.
  *If not, why:* The integration suite creates EKS clusters and AWS secrets, which I did not run. The change is expected to make test 11 converge in seconds rather than at the reconciler's five-minute tick.
- [x] I have updated integration tests (if needed)
  *If not, why:* No integration test change is needed; the existing test 11 assertion becomes fast rather than marginal.
  *If not, why:*
  *If not, why:*
- [x] I have updated README/documentation (if needed)
  *If not, why:* No documented behavior changes.
  *If not, why:* No breaking changes. Object versions reported for mounts using `failoverRegion` change from the failover region's version to the primary's, which causes one extra fetch on the first republish after upgrade.

---

## Reviewer Checklist

---


provider/parameter_store_provider.go
```diff
@@ -81,7 +82,14 @@ func (p *ParameterStoreProvider) fetchParameterStoreValue(
 ) (values []*SecretValue, err error) {
 
 	for _, client := range p.clients {
-		batchValues, err := p.fetchParameterStoreBatch(ctx, client, batchDescriptors, curMap)
+		// Only the client that supplies the mounted values reports versions; a
+		// later client writes into a copy that is discarded.
+		verMap := curMap
+		if len(values) > 0 {
+			verMap = maps.Clone(curMap)
```
bob2681312, line 89: Would doing this add a lot of latency when retrieving secrets with a failover region? If yes, is there anyway you could avoid just copying the map? 

An idea I had was to pass in a boolean specifying if the map is read-only, and only make edits to the map in fetchParameterStoreBatch depending on the boolean, but this could make the code confusing and harder to read.

provider/secrets_manager_provider.go
```diff
@@ -76,7 +77,15 @@ func (p *SecretsManagerProvider) fetchSecretManagerValue(
 ) (value []*SecretValue, err error) {
 
 	for _, client := range p.clients {
-		secretVal, err := p.fetchSecretManagerValueWithClient(ctx, client, descriptor, curMap)
+		// Only the client that supplies the mounted value reports versions. A
+		// later client still needs the current versions to decide whether to
+		// fetch, so it gets a copy whose writes are discarded.
+		verMap := curMap
+		if len(value) > 0 {
+			verMap = maps.Clone(curMap)
```
bob2681312, line 85: Same thing here

server/server_version_report_test.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if ver.Version != exp {
+			t.Errorf("%s: Expected %s version %s got %s", tst.testName, ver.Id, exp, ver.Version)
+		}
+		delete(expVersions, ver.Id)
+	}
+	if len(expVersions) != 0 {
+		t.Errorf("%s: Missing object versions %v", tst.testName, expVersions)
+	}
+}
```
bob2681312, line 76: nit: Maybe also add a test that when the primary fails, make sure the version matches that of the backup

